PR #186, opened 2015-01-07
Rather than calling getMap when initializing the controller for the Coun...

mbucknell:
...try Selection StepController, use initMap. The application comes to this step from a route that no longer contains the existing map object's dom element. By calling initMap, the map object is recreated and this avoids problems when displaying the map initially. Removed code that was added to try to work around this issue that is no longer needed.

### DIFF
--- a/nwc/src/main/webapp/client/nwc/workflows/waterBudget/waterBudgetControllers.js
+++ b/nwc/src/main/webapp/client/nwc/workflows/waterBudget/waterBudgetControllers.js
@@ -60,14 +60,14 @@ waterBudgetControllers.controller('PlotData', ['$scope', '$state', 'StoredState'
             	var countyVectorLayer = new OpenLayers.Layer.Vector("Simple Geometry County", {
                 	style: layerStyle
             	});
-            	
+
             	var countyFeature = new OpenLayers.Feature.Vector(StoredState.countyFeature.geometry);
             	countyVectorLayer.addFeatures([hucFeature.clone(), countyFeature]);
 
     			$scope.countyLayer = [countyVectorLayer];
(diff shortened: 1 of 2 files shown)


nwc/src/main/webapp/client/nwc/workflows/waterBudget/waterBudgetControllers.js
```diff
@@ -263,13 +263,12 @@ waterBudgetControllers.controller('SelectCounty', ['$scope', 'StoredState', 'Com
         $scope.CommonState = CommonState;
         $scope.isCountySelectionPage = true;
 
-        var map = WaterBudgetMap.getMap();
+        var map = WaterBudgetMap.initMap();
+		map.zoomToExtent(StoredState.mapExtent, true);
```
jiwalker-usgs: different indentation here

mbucknell:
There are a lot of indentation problems. I have my netbeans set up for tabs and there are a lot of spaces. Started to fix, but it seemed like it would make the diff very big for what is a small change.

jiwalker-usgs:
Yeah, no problem really, just triggered my comment finger.

jiwalker-usgs:
:+1: I'm thinking we may want to have getMap() check whether it has the DOM it needs and reinit itself if it doesn't rather than having two calls floating around.  Something to think about.

mbucknell:
The getMap call is within a factory. Would it be weird to be querying the dom from there?

mbucknell:
I thought about it more. The getMap call is working. The problem is that the dom element where the original initMap call was made has been removed as another "page" was loaded. When we go to the "page" for the map showing counties it is using the same named dom element as the first page, so the check in getMap works and returns the map. This is what was causing the issues in this ticket. Reinitializing the map is what needed to be done.

PR #191, opened 2015-02-26
Watersmart 490

mbucknell:
All functionality for the waterbudget page is in place. I will be doing some code cleanup and possibly adding tests too.

The idea for BaseSelectMapView is much of the code can be directly reused for the other workflow selection maps.


mamcderm-usgs:
:+1:

PR #192, opened 2015-02-27
Added test for the BaseSelectMapView and the WaterBudgetMapView. Fixed s...

mbucknell:
...everal issues as a result. Added more code to mapUtils.js to make stubbing out map initialization code simpler.

This was mostly reorganization and writing tests, although one important change is I moved the instantiation of a view's model into the initialization procedure. One of the view properties is Model which is the model's class.


mamcderm-usgs:
:+1:

PR #193, opened 2015-03-02
Watersmart 491

mbucknell:
Things to note. I started using handlebars partials. I created a subdirectory in templates for the partials and added code to templateLoader to load the partials by name.

I also basically picked up the Conversion service and put it in js/utils and took out the "angular". There were tests already written for this service so I moved those too. As we convert code, we should look in test/old_specs to see if tests already exist that can be easily modified and put into test/specs.


mamcderm-usgs:
Looks good to me. :+1:

PR #195, opened 2015-03-04
Watersmart 492

mbucknell:
Things to note:

I changed the joinProtocols call in AquaticBiologyMapSelectView so that the GetFeature control would only fire one featuresselected event per operation rather than three.

I will be adding jasmine-jquery plugin to make loading templates and testing jquery selectors simpler during tests. I will also be refactoring the water budget and streamflow stats pages to update the view's html when handing model updates. I was only updating the map, but the buttons on the page reflect the state of the control's and those need to be updated too.


mamcderm-usgs:
Looks good :+1:

PR #197, opened 2015-03-05
Watersmart 493

mbucknell:
In addition to adding a modal warning, I instantiated the view's models in the router so that state can be restored when using the back button. Also added code to save the map extent so it too can be restored.


mamcderm-usgs:
:+1:

PR #199, opened 2015-03-09
Watersmart 474

mbucknell:
Using the search data service rather then the javascript search API out of txpub. I'm going to go ahead and merge this so Blodgett can see how it works.  There will be some clean-up that needs to be done including put the search data service endpoint into context.xml.


mamcderm-usgs:
:+1:

PR #200, opened 2015-03-09
Watersmart 474

mbucknell:
Addressed comments from D. Blodgett and also made the searchService endpoint configurable. Added additional tests.


mamcderm-usgs:
:+1:

PR #202, opened 2015-03-11
Watersmart 494

mamcderm-usgs:
File download is working but I need to update the file naming after I get the inset map working.


nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -200,14 +200,14 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
 	},
 
 	downloadEvapotranspiration : function() {
-//		var blob = new Blob([NWC.util.dataSeriesStore.eta.toCSV(true)], {type:'text/csv'});
-//		saveAs(blob, getHucFilename('eta'));
+		var blob = new Blob([NWC.util.DataSeriesStore.eta.toCSV(true)], {type:'text/csv'});
```
mbucknell: I don't think that toCSV takes a boolean parameter so that should be removed.

mbucknell:
Mark, I'm going to go ahead and merge.

PR #203, opened 2015-03-12
Watersmart 498

mbucknell:


mbucknell:
Mark, this has the code that you will use to do the inset map for the water budget huc data page. You will use NWC.util.mapUtils.createHucFeatureLayer rather than the one I'm using in StreamflowStatsHucDataView.js. Otherwise the code should be similar.

Please review and let me know. I will merge it after your review.

mamcderm-usgs:
Looks good :+1:

PR #205, opened 2015-03-12
Watersmart 494

mamcderm-usgs:
Pretty close with this merge request but still small problem with file download naming.  It doesn't break, just not working quite right yet.  However, requesting merge to avoid to much delay.  This request includes:

- jasmine spec files for DataSeries and SosResponseParser
- inset map
- miscellaneous fixes, cleanup


nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -1,244 +1,244 @@
...
+						});
+					}
+					csvHeader += this.metadata.seriesLabels.map(function(label) {
+						return createSeriesLabel(label);
+					}).join(",") + "\r\n";
+					var csvValues = "";
+					this.data.each(function(row) {
+						csvValues += row.join(",") + "\r\n";
+					});
+//					return encodeURIComponent(csvHeader + csvValues);
```
mbucknell: Take out commented out code.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -1,233 +1,230 @@
...
+		this.map.render(this.insetMapDiv);
+	},
+	
+	huc12 : null,
+	hu12Name : null,
+
+	initialize : function(options) {
+		
+		if (options.hucId) {
+			this.hucId = options.hucId;
```
mbucknell: So this.hucID and this.huc12 should be the same. One of them should be eliminated.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -1,233 +1,230 @@
...
+		var hucLayer = NWC.util.mapUtils.createHucFeatureLayer(this.hucId);
+
+		hucLayer.events.on({
+			featureadded: function(event){
+				this.map.zoomToExtent(this.getDataExtent());
+				
+				this.huc12 = event.feature.attributes.HUC_12;
+				this.hu12Name = event.feature.attributes.HU_12_NAME;
+				$('#huc-id').html(event.feature.attributes.HUC_12);
+				$('#huc-name').html(event.feature.attributes.HU_10_NAME);
```
mbucknell: Should this be the HU_10_NAME? Might want to verify with Blodgett. Is there are reaso to store the hu12Name then?

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -1,233 +1,230 @@
...
+					labeledResponses[label] = labeledDataSeries;
+					d.resolve();
+				},
+				context : sourceId,
+				dataType : "xml",
+				error : function() {
+                    //@todo - setup app level error handling
+                    var errorMessage = 'error retrieving time series data';
+                    alert(errorMessage);
+                    $log.error(errorMessage);
```
mbucknell: Eliminate the $log. These won't work anyways because that's an angular method.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -1,233 +1,230 @@
...
+
+	getHucFilename : function (series) {
+		var filename = series + '_data.csv';
+        if (this.hu12Name || this.huc12) {
+        	filename = this.buildName(this.hu12Name, this.huc12, series);
+        }
+		return filename;
+	},
+
+	buildName : function(selectionName, selectionId, series) {
```
mbucknell: I'm not sure what you need for filenames. You might consider righting a test for this so it's well documented.

mbucknell:
General comments: are you reformatting the files? I'm wondering why three of the files show changes for the whole file.

I verified with Blodgett that we should be displaying the HUC_12_Name, not HUC 10. The old code is wrong.

Good to merge. You can address any comments with the next set of changes. :shipit:

PR #206, opened 2015-03-13
Fixed bug where I was using the HU_12_DS attribute rather than HUC_12 to...

mbucknell:
... get the huc id for the next step.


mbucknell:
Thanks to Mark for catching this error!

PR #207, opened 2015-03-13
Watersmart 498

mbucknell:


mbucknell:
The inset map now works correctly and adding the calculating stats portion of the page. Still to do is the plotting portion of the page.

mamcderm-usgs:
looks good :+1:

mbucknell:
Looks like I didn't merge from before ... my bad. The new code added handles plotting the streamflow data which should be very similar to the water budget huc data page view.

mbucknell:
@mamcderm-usgs, could you please review when you have a chance.

mamcderm-usgs:
Looks good :+1:

PR #208, opened 2015-03-13
Watersmart 499

kmschoep-usgs:
Selected biodata sites display with checkboxes.


mbucknell:
If you want to split the pages templates into pieces, consider using Handlerbar partials. I'm using partials in the various SelectMapView's. It may also make sense to just have the one template for the page. Then there is only a single call to Handlebars to render a template for the page.

kmschoep-usgs:
I'll have to review handlebars a bit more, when I tried keeping it all on one page, the entire template got duplicated in the sites-table-div.

PR #210, opened 2015-03-16
show all types of lists: gages, hucs and sites.  There are checkboxes in...

kmschoep-usgs:
... site list, but they are not bound to anything yet.


mbucknell:
Looks good!

PR #211, opened 2015-03-16
Watersmart 499

kmschoep-usgs:
get checkboxes working, both individually and as "select all" to populate array of biodata site ID's.


nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -15,5 +21,38 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
...
+                if ($cb.is(':checked')) {
+                    this.model.set({ 
+                    'selected' : this.model.get('selected').concat(name)
+                    });
+                } else {
+                    var selected = this.model.get('selected');
+                    var index = selected.indexOf(name);
+                        if (index > -1) {
+                        selected.splice(index, 1);
+                        }
```
mbucknell: I think this.model.set needs to be called in the else part.

mbucknell:
:+1:

PR #213, opened 2015-03-17
send selected huc to streamflow stats page

kmschoep-usgs:


nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -11,16 +11,26 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
...
         },
 
         initialize : function() {
+            var tempHucs = [];
+            var filteredHucs = [];
             this.context.biodataSites = this.model.get('sites');
             this.context.gages = this.model.get('gages');
-            this.context.hucs = this.model.get('hucs');
+            tempHucs = this.model.get('hucs');
+            $.each(tempHucs, function(key, huc) {
```
mbucknell: I think this filtering should occur in AquaticBiologyMapView. You can use Sugar's filter function prior to using the map function.

nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -11,16 +11,26 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
...
+            if ( parseFloat(huc.DRAIN_SQKM) < 2000 ) {
+                    filteredHucs.push(huc);
+                }
+            });
+            this.context.hucs = filteredHucs;
             $('#sites-table-div').html({biodataSites : this.model.get('sites')});
             $('#gages-table-div').html({gages : this.model.get('gages')});
-            $('#hucs-table-div').html({hucs : this.model.get('hucs')});
+            $('#hucs-table-div').html({hucs : this.context.hucs});
```
mbucknell: I don't think these three lines do anything. At this point the table isn't in the DOM because the template hasn't been rendered.

kmschoep-usgs:
ready for another review.

mbucknell:
:+1:

PR #214, opened 2015-03-18
Watersmart 497

mbucknell:
Created a BaseStreamFlowStatsDataView since both the gage and huc view's calculate, display, and download stats. These view's now extend BaseStreamFlowStatsDataView.

Also brought in with very few changes RdbParser.


kmschoep-usgs:
Nice!  I like the new baseview, and I've got some ideas for tests for my part.

:thumbsup:

PR #215, opened 2015-03-18
Watersmart 494

mamcderm-usgs:
This should be close...


nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -29,39 +29,58 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
...
+		this.hucLayer.events.on({
 			featureadded: function(event){
-				this.map.zoomToExtent(this.getDataExtent());
-				
-				$('#huc-id').html(event.feature.attributes.HUC_12);
+				this.hucName = event.feature.attributes.HU_12_NAME;
+				this.map.zoomToExtent(this.hucLayer.getDataExtent());
+
 				$('#huc-name').html(event.feature.attributes.HU_12_NAME);
+				this.$(".evapotranspiration-download-button").removeAttr("disabled");
```
mbucknell: You should use .prop rather than removeAttr, e.g. this.$('el').prop('disabled', false); . The attribute is for initial rendering of the html but after that you'll need to use .prop to set properties such as 'disabled', 'checked', and 'selected'.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -137,66 +156,65 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
 	},
 
 	toggleMetricLegend : function() {
-		this.$(".customary-button").removeAttr("disabled");
```
mbucknell: same comment as above

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -137,66 +156,65 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
...
+			this.plotPTandETaData(this.MONTHLY, this.METRIC);
 		}
 	},
 
 	toggleCustomaryLegend : function() {
-		this.$(".metric-button").removeAttr("disabled");			
-		this.$(".customary-button").attr("disabled","disabled");			
-		if (this.$(".daily-button").attr("disabled")) {
-			this.plotPTandETaData(this.DAILY, this.CUSTOMARY);			
+		$(".metric-button").removeAttr("disabled");
```
mbucknell: same comment as above.

nwc/src/test/javascript/specs/WaterBudgetHucDataView_spec.js
```diff
@@ -1,58 +1,122 @@
...
+		expect(testView.events['click .customary-button']).toBeDefined();
+		expect(testView.events['click .monthly-button']).toBeDefined();
+		expect(testView.events['click .daily-button']).toBeDefined();
+		expect(testView.events['click .evapotranspiration-download-button']).toBeDefined();
+		expect(testView.events['click .precipitation-download-button']).toBeDefined();
+
+		//plot buttons exist and get set with the proper disabled attribute
+		testView.plotPTandETaData = jasmine.createSpy('plotPTandETaDataSpy')
+		testView.toggleMetricLegend();
+		expect($('.customary-button').attr('disabled')).toBeUndefined();
```
mbucknell: When you change to using .prop, you'll need to change the test to use .prop to return the state.

mbucknell:
Looks good other than the use of removeAttr and attr rather than .prop. This is one of those things that it may work in one browser to use attr but won't in another.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
mamcderm-usgs: Mary - should I remove "this" from line 64?  I'm not seeing or remembering how jquery works with "this".

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
mamcderm-usgs: should I also set "disabled" using prop....or is attr better?

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
mbucknell: The this is not needed here.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
mbucknell: Yes. 'disabled' should be set using  .prop

PR #216, opened 2015-03-19
Added a view for the modeled info page and then corrected the header in ...

mbucknell:
...the modeled data download file. Also fixed the link on the Streamflow Stats map page. Fixed the header on the gage download.


mbucknell:
I know this looks like a lot but it is actually mostly adding a text documentation page that we missed. I'm going to go ahead and merge.

PR #218, opened 2015-03-20
Watersmart 495

mbucknell:
For the moment, I have the route for the county page using the WaterBudgetHucDataView. This will have to change. Mark, you should also be able to use hucCountiesIntersector.js to display the percentages on the huc/county data page.

### DIFF
--- a/nwc/src/main/webapp/js/controller/NWCRouter.js
+++ b/nwc/src/main/webapp/js/controller/NWCRouter.js
@@ -18,6 +18,7 @@ NWC.controller.NWCRouter = Backbone.Router.extend({
 		'' : 'home',
 		'home' : 'home',
 		'waterbudget/huc/:huc' : 'waterbudgetHucData',
+		'waterbudget/huc/:huc/county/:fips' : 'waterbudgetHucCountyData',
 		'waterbudget' : 'waterbudget',
 		'streamflow-stats' : 'streamflowStats',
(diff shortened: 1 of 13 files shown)


nwc/src/main/webapp/js/controller/NWCRouter.js
```diff
@@ -45,6 +46,14 @@ NWC.controller.NWCRouter = Backbone.Router.extend({
 			insetMapDiv : 'huc-inset'});
 	},
 
+	waterbudgetHucCountyData : function(huc, fips) {
+		this.showView(NWC.view.WaterBudgetHucDataView, {
+			hucId : huc,
+			fips : fips,
+			insetMapDiv : 'huc-inset'
```
mamcderm-usgs: will we also need a "insetCountyMapDiv : 'county-inset'"
actually, I've been meaning to ask, why do we need to set these map divs?

mamcderm-usgs:
Looks good :+1:

nwc/src/main/webapp/js/controller/NWCRouter.js
mbucknell: We will need a county inset div. We don't really need to set these, we could just default the string value in the view. I started it this way out of habitat. When I've written code to create a map, I almost always pass in the div id. Many times, its because that piece of code is reused and/or it makes testing easier. Neither of those really apply here. You'll notice that I didn't pass in the div for the county selection map.

PR #219, opened 2015-03-23
Updated eta SOS filename to HUC12_eta.nc

mbucknell:
Fixed as requested


mbucknell:
Only changed the file under js/utils because the other will be removed during cleanup.

PR #221, opened 2015-03-24
Need to create a new feature for the huc layer when passing instantiatin...

mbucknell:
...g the HucCountMapView.


mbucknell:
@mamcderm-usgs, I believe you will need to do something similar on the county-huc page when adding the hucLayer to the county inset. The feature actually has a reference to the layer it's been added to. And that reference is in the geometry property. That's why you have to clone the geometry when creating a new feature.vector object.

You can either merge this or integrate it into your changes on your branch.  If you decide to integrate it yourself, please close the pull request.

mamcderm-usgs:
I integrated this into my branch, so closing.

PR #224, opened 2015-03-27
Watersmart 496

mamcderm-usgs:
- completed tests for watersmart-496
- some cleanup made while working on tests

### DIFF
--- a/nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
+++ b/nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
@@ -36,7 +36,6 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
 
 		this.context.hucId = options.hucId;
 		this.hucId = options.hucId;
-		this.fips = options.fips;
 		this.insetHucMapDiv = options.insetHucMapDiv;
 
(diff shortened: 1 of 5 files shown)


nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -156,7 +155,6 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
 	},
 
 	displayCountyMap : function() {
-		$('#county-selection-div').show();
```
mbucknell: Don't we still need this line?

mbucknell:
:+1: , except for the comment above.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
mamcderm-usgs: I believe this is superfluous, unless I'm missing something.  It's just an empty div until you click the "show counties" button.  I'm going to go ahead and merge but I wouldn't rule out needing to put this back at some point.

mbucknell: Ok ... You are correct. Merge away!

PR #225, opened 2015-03-27
Changes for counties and HUCs in postgis

dblodgett-usgs:
I'll have another round of changes for the national HUCs once those are all loaded up. Wanted to get this into the canonical code before more work happens though.

### DIFF
--- a/nwc/src/main/webapp/js/view/HucCountyMapView.js
+++ b/nwc/src/main/webapp/js/view/HucCountyMapView.js
@@ -57,7 +57,7 @@ NWC.view.HucCountyMapView = NWC.view.BaseView.extend({
 		// Set up control
 		var selectControl = new OpenLayers.Control.SelectFeature(this.countiesLayer, {
 			onSelect : (function(feature) {
-				this.router.navigate('/waterbudget/huc/' + options.hucFeature.attributes.HUC_12 + '/county/' + feature.attributes.FIPS, {
+				this.router.navigate('/waterbudget/huc/' + options.hucFeature.attributes.HUC_12 + '/county/' + feature.attributes.fips, {
 					trigger : true
 				});
 			}).bind(this)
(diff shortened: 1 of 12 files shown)


nwc/src/main/webapp/js/view/HucCountyMapView.js
```diff
@@ -57,7 +57,7 @@ NWC.view.HucCountyMapView = NWC.view.BaseView.extend({
 		// Set up control
 		var selectControl = new OpenLayers.Control.SelectFeature(this.countiesLayer, {
 			onSelect : (function(feature) {
-				this.router.navigate('/waterbudget/huc/' + options.hucFeature.attributes.HUC_12 + '/county/' + feature.attributes.FIPS, {
+				this.router.navigate('/waterbudget/huc/' + options.hucFeature.attributes.HUC_12 + '/county/' + feature.attributes.fips, {
```
mbucknell: I think you may need to lower case options.hucFeature.attributes.HUC_12. The options.hucFeature is the hucFeature displayed in the inset map on the water budget data page.

nwc/src/main/webapp/templates/aquaticBiologySelectFeatures.html
```diff
@@ -77,8 +77,8 @@
                                 </thead>
                                 <tbody id="hucList">
                                     {{#each hucs}}
-                                    <tr id="{{HUC12}}">
-                                        <td class="xs-col-2">{{HUC12}}</td>
+                                    <tr id="{{huc12}}">
+                                        <td class="xs-col-2">{{huc12}}</td>
                                         <td class="xs-col-4">{{HU_12_NAME}}</td>
```
mbucknell: Ditto here. These are tied to the hucFeature retrieved.

mbucknell:
Should double check places where I commented. Merge when you think its ready.

nwc/src/main/webapp/js/view/HucCountyMapView.js
dblodgett-usgs: Yep, I just hadn't done the national wbd yet.

nwc/src/main/webapp/templates/aquaticBiologySelectFeatures.html
dblodgett-usgs: Yeah... it got confusing with the regional and national hucs. Might be something to clean up and put into variables. We could use this pull request as reference for that.

mbucknell: The attribute names are different in the two different layers and this just reflects that. It might be better to try to get the attribute names the same in the layers rather than deal with it in the client.

PR #227, opened 2015-03-31
watersmart-507; initial review

mamcderm-usgs:
This should work but I can't connect to the web service to test locally.  This will line will need to be added to context.xml to work locally:

<Environment name="nwc.endpoint.sciencebase" override="true" type="java.lang.String" value="https://beta.sciencebase.gov"/>

### DIFF
--- a/nwc/src/main/webapp/js/view/DataDiscoveryView.js
+++ b/nwc/src/main/webapp/js/view/DataDiscoveryView.js
@@ -6,41 +6,24 @@ NWC.view.DataDiscoveryView = NWC.view.BaseView.extend({
 	templateName : 'dataDiscovery',
 
 	initialize : function(options) {
-		if (!Object.has(this, 'context')) {
-			this.context = {};
-		}
-		
-		var _this = this;
-		var projectDataPromise = this.getProjectData(_this);
-		$.when(projectDataPromise).done(function() {
 
-			// call superclass initialize to do default initialize
-			// (includes render)
-			NWC.view.BaseView.prototype.initialize.apply(_this, arguments);			
-		});
-		
+		// call superclass initialize to do default initialize
+		// (includes render)
+		NWC.view.BaseView.prototype.initialize.apply(this, arguments);			
+
+		this.getProjectData();
 	},
 
 	/**
 	 * This makes a Web service call to get project data for display
 	 */
-	getProjectData: function(_this) {
-		var d = $.Deferred();
+	getProjectData: function() {
 		var projects = [];
 		$.ajax({
 			url : CONFIG.endpoint.direct.sciencebase + '/catalog/items?facetTermLevelLimit=false&q=&community=National+Water+Census&filter0=browseCategory%3DProject&max=100&format=json',
 			dataType : "json",
 			success: function(data) {
-				for (var i = 0; i < data.total; i++) {
-					if (data.items[i].summary) {
-						projects.push({title: data.items[i].title, summary: data.items[i].summary});
-					}
-					else {
-						projects.push({title: data.items[i].title, summary: 'Summary unavailable'});				
-					}
-				}
-				_this.context.projectList = projects;
-				d.resolve();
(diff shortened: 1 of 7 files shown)


nwc/src/main/webapp/js/view/DataDiscoveryView.js
```diff
@@ -1,9 +1,51 @@
...
+			$('#all-projects').append('<h5>' + data.items[i].title + '</h5>');
+			if (data.items[i].summary) {
+				$('#all-projects').append('<p>' + data.items[i].summary + '</p>');				
+			}
+			else {
+				$('#all-projects').append('<p>Summary unavailable</p>');								
+			}
+			$('#all-projects').append('<div class="row spacer"></div>');
+		}
+	}
```
mbucknell: Rather than use jquery to create the DOM elements, I would prefer that Handlebars be used. You can create a template just for this piece of the page. As an example, look at https://github.com/USGS-CIDA/nwc/blob/master/nwc/src/main/webapp/js/view/BaseStreamflowStatsDataView.js in calculateStats. You can use the built-in helper "if" to do the if-else statement right in the template.

nwc/src/main/webapp/js/view/DataDiscoveryView.js
```diff
@@ -1,9 +1,55 @@
...
+			success: function(data) {
+				for (var i = 0; i < data.total; i++) {
+					if (data.items[i].summary) {
+						projects.push({title: data.items[i].title, summary: data.items[i].summary});
+					}
+					else {
+						projects.push({title: data.items[i].title, summary: 'Summary unavailable'});				
+					}
+				}
+				_this.context.projectList = projects;
```
mbucknell: Rather than make a new object, projects to pass to the context, you could just pass data.items. This way other fields are available to the template and the javascript does not have to change to show those fields.

nwc/src/main/webapp/js/view/DataDiscoveryView.js
```diff
@@ -1,9 +1,55 @@
...
+			this.context = {};
+		}
+		
+		var _this = this;
+		var projectDataPromise = this.getProjectData(_this);
+		$.when(projectDataPromise).done(function() {
+
+			// call superclass initialize to do default initialize
+			// (includes render)
+			NWC.view.BaseView.prototype.initialize.apply(_this, arguments);
```
mbucknell: This will delay rendering anything on the page until the sciencebase call returns. Another way to do this would be to render the page before the promise is done, and then render what is now a partial (would need to change to a template) when resolving the promise. Just food for thought.

mamcderm-usgs: I struggled with what to do here.  If I change the partial to a template, don't I also need to make a View for it or is there a simpler way to render a template within a template?  Creating another View seemed making it over-complicated.  But I also don't like waiting to render until the promise is done.  I expect it's ok this way for now, but I'd also like to know the best way to handle this.  Thanks.

nwc/src/main/webapp/js/view/DataDiscoveryView.js
mamcderm-usgs: Of course, much better, will do this!  Thanks.

nwc/src/main/webapp/js/view/DataDiscoveryView.js
mbucknell: You do not need to make a separate view. You can render a template anywhere and at anytime, it's just a string that gets added to the DOM. If you look at line 58 in https://github.com/USGS-CIDA/nwc/blob/master/nwc/src/main/webapp/js/view/BaseStreamflowStatsDataView.js that's exactly what I am doing there. I use the NWC.templates.getTemplate to retrieve the template function and then call the function with the context I need (the object with a streamflowStatistics property). The result of this call is a string which I insert into the dom with the .html() function.

mbucknell:
Yes ... that's way to render a template. It's good to go! :+1:

PR #234, opened 2015-04-03
Watersmart 510

mbucknell:
Lots of little changes. Some things to note:

Changed the implementation of radio buttons on the data views to be the same as is used on the map views. 

Factored out the county data portion of the water budget huc/county data page template into a separate template that is only rendered when we are showing county data. 

Lots moved around on the data pages. There is now two unit radio button sets on the county huc data page, one for the county water use plot and one for the huc plot.


mamcderm-usgs:
wow, quite a refactor of a refactor:-)  toggle refactor is awesome!  :+1:

PR #235, opened 2015-04-03
Watersmart 508

mamcderm-usgs:


mamcderm-usgs:
Mary - This is a version of the test for DataDiscoveryView using sinon.fakeserver.  Is it ok to keep both versions for reference or would you rather only one?

mbucknell:
There should only be one version. I like the version using sinon.fakeserver so that's the one I would recommend keeping.

PR #241, opened 2015-04-20
Watersmart 514

mbucknell:


mbucknell:
@dblodgett-usgs, look in js/utils/CountyWaterUseProperties.js for the colors associated with each long name. You can also adjust the fill opacity in js/utils/waterUsageChart.js

PR #245, opened 2015-05-13
Watersmart 517

kmschoep-usgs:
@mbucknell, can you review at your leisure?

### DIFF
--- a/nwc/src/main/webapp/js/view/AquaticBiologyMapView.js
+++ b/nwc/src/main/webapp/js/view/AquaticBiologyMapView.js
@@ -167,7 +167,6 @@ NWC.view.AquaticBiologyMapView = NWC.view.BaseSelectMapView.extend({
                                     return n.attributes.drain_sqkm < 2000;
                                 });
 				this.aquaticBiologyFeaturesModel.set({
-					//sites : siteFeatures.map(function(f) { return f.attributes; }),
                                         sites : siteFeatures,
 					gages : gageFeatures.map(function(f) { return f.attributes; }),
 					hucs : filteredHucFeatures.map(function(f) { return f.attributes; })
--- a/nwc/src/main/webapp/js/view/AquaticBiologyMapView.js
+++ b/nwc/src/main/webapp/js/view/AquaticBiologyMapView.js
@@ -68,13 +68,13 @@ NWC.view.AquaticBiologyMapView = NWC.view.BaseSelectMapView.extend({
 			featureType: 'SiteInfo',
 			featureNS: 'http://cida.usgs.gov/BioData',
 			srsName: 'EPSG:900913',
-			propertyNames: ['SiteNumber', 'SiteName']
+			propertyNames: ['SiteNumber', 'SiteName', 'the_geom']
 		});
 
 		var gageProtocol = OpenLayers.Protocol.WFS.fromWMSLayer(this.gageFeatureLayer, {
 			url : CONFIG.endpoint.geoserver + "wfs",
 			srsName : "EPSG:3857",
-			propertyNames: ["STAID","STANAME","DRAIN_SQKM"]
+			propertyNames: ["STAID","STANAME","DRAIN_SQKM","the_geom"]
 		});
 
 		var hucProtocol = OpenLayers.Protocol.WFS.fromWMSLayer(this.hucLayer, {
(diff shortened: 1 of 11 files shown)


nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -108,6 +113,16 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
                 $cb = $(e.currentTarget);
                 gageID = $cb.attr('id');
                 this.router.navigate('/streamflow-stats/gage/' + gageID, {trigger : true});
-            } 
+            },
+	    
+	    displayMap : function() {	
+		this.biodataGageMapView = new NWC.view.BiodataGageMapView({
+			mapDiv : 'biodata-gage-selection-map',
+			//biodataFeature : mapFeatures,
```
mbucknell: Take out commented out code.

nwc/src/main/webapp/templates/biodataGageMap.html
```diff
@@ -0,0 +1,2 @@
+<div class="spacer">Biodata and NWIS gages</div>
```
mbucknell: You don't have to use a template within a view. I think in this case, I would put this html in aquaticBiologySelectFeatures.html. The new BiodataGageMapView doesn't have to extend NWC.BaseView, it could just extend Backbone.View. An advantage to putting in right in aquaticBiologySelectFeatures is that everything is within that single template.

nwc/src/main/webapp/templates/biodataGageMap.html
```diff
@@ -0,0 +1,2 @@
...
+<div id="biodata-gage-selection-map" style="height:300px;border-style:solid;border-width:1px;"></div>
```
mbucknell: Eventually you need to put the style in custom.css.

nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -108,6 +113,16 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
                 $cb = $(e.currentTarget);
                 gageID = $cb.attr('id');
                 this.router.navigate('/streamflow-stats/gage/' + gageID, {trigger : true});
-            } 
+            },
+	    
+	    displayMap : function() {
```
mbucknell: I have been putting an underscore in front of functions that are not intended to be used outside of the view.

nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -34,13 +35,17 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
                     this.model.set({ 
                     'selected' : this.model.get('selected').concat(name)
                     });
+		    var selectedSite = this.biodataGageMapView.getSelectedSiteFeature(name);
+		    this.biodataGageMapView.highlightSite(selectedSite);
```
mbucknell: You could combine these into a single function in biodataGageMapView.highlightSite(name)

nwc/src/main/webapp/js/view/AquaticBiologyMapView.js
```diff
@@ -167,7 +167,7 @@ NWC.view.AquaticBiologyMapView = NWC.view.BaseSelectMapView.extend({
                                     return n.attributes.drain_sqkm < 2000;
                                 });
 				this.aquaticBiologyFeaturesModel.set({
-					sites : siteFeatures.map(function(f) { return f.attributes; }),
+                                        sites : siteFeatures,
```
mbucknell: So you didn't need to clone the geometry? Now that I think about it, the reason I had to do it is I was cloning a feature that was in the other map on the page. So I would create the new county feature and the county would go away from the other page.

Bottom line is that this looks like it's working so don't make it more complicated. Just keep the issue in mind if there is weirdness later.

nwc/src/main/webapp/js/view/BiodataGageMapView.js
```diff
@@ -0,0 +1,116 @@
...
+		this.map.render(this.mapDiv);
+		this.map.zoomToExtent(this.biodataFeatureLayer.getDataExtent());
+		return this;
+	},
+	
+	highlightSite : function(selected_item) {
+	    this.selectControl.select(selected_item[0]);
+	},
+	
+	checkSite : function(feature) {
```
mbucknell: You shouldn't be referencing something in the parent view. Would be better to pass in the checkSite and unCheckSite functions when you initialize the view.

mbucknell:
Some things to address as you go along. Merge if you want.

kmschoep-usgs:
It does currently highlight a gage row.  If/when that functionality is removed, I can rename the function.

mbucknell:
I must be missing something. The method highlightGageRow just shows an alert. How does it highlight a row?

kmschoep-usgs:
Sorry, the commits are behind.  Those functions now do highlight/unhighlight gage rows on my local.

mbucknell:
Sorry about reviewing this after merging. I don't think I got a notification for the further review that you wanted. If you want me to look at pull request and guarentee that I get an email, but an '@' and my github login (mbucknell) in the message.

PR #246, opened 2015-05-20
Watersmart 517

kmschoep-usgs:
@mbucknell , I got the map selections working!  Can you please review when you have time?

### DIFF
--- a/nwc/src/main/webapp/js/model/AquaticBiologyFeaturesModel.js
+++ b/nwc/src/main/webapp/js/model/AquaticBiologyFeaturesModel.js
@@ -8,7 +8,20 @@ NWC.model.AquaticBiologyFeaturesModel =  Backbone.Model.extend({
             sites : [],
             gages : [],
             hucs : [],
-            selected : []
-        }
+            selected : [],
+            pairs : []
+        },
+	associatePairs : function(s, g, action) {
+        this.updPairs = this.get('pairs').clone();
+	    if (action === 'add') {
+            this.updPairs.add({site_id: s, gage_id : g});
+	    };
+	    if (action === 'remove') {
+            this.updPairs.remove(function(n) {
+	        return n['site_id'] === s && n['gage_id'] === g;
+            });
+	    };
+        this.set({pairs : this.updPairs});
+	}
                         
 });
--- a/nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
+++ b/nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
@@ -126,11 +126,13 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
 	    highlightGageRow : function(feature) {
 		var selectedLayer = feature.layer.name;
 		alert('Selected ' + selectedLayer);
+		$('#' + feature.attributes.STAID).addClass('gage-selected');
 	    },
 	
 	    unHighlightGageRow : function(feature) {
 		var selectedLayer = feature.layer.name;
 		alert('Unselected ' + selectedLayer);
+		$('#' + feature.attributes.STAID).removeClass('gage-selected');
 	    }
         
 });
--- a/nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
+++ b/nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
@@ -1,3 +1,5 @@
+/* global this */
+
 var NWC = NWC || {};
 
 NWC.view = NWC.view || {};
@@ -13,13 +15,17 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
             'click #allSelected' : 'selectAll',
             'click #biodata-form-button' : 'sitesDoc',
             'click #hucs-table-div tr' : 'onHucSelect',
-            'click #gages-table-div tr' : 'onGageSelect'
+            'click #gages-table-div tr' : 'onGageSelect',
+            'click .dismiss-btn' : 'removePair'
         },
 
         initialize : function() {
             this.context.biodataSites = this.model.get('sites');
             this.context.gages = this.model.get('gages');
             this.context.hucs = this.model.get('hucs');
+            this.context.pairs = this.model.get('pairs');
+            this.listenTo(this.model, 'change:pairs', this.displayPairList);
+            this.listTemplate = Handlebars.compile('{{#each pairs}}<div>Site ID: {{site_id}}, Gage ID: {{gage_id}}<button title="Remove Pair" id="{{site_id}}" name="{{gage_id}}" class="dismiss-btn">x</button></div>{{/each}}');
             NWC.view.BaseView.prototype.initialize.apply(this, arguments);
             this._displayMap();
         },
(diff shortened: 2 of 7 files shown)


nwc/src/main/webapp/js/model/AquaticBiologyFeaturesModel.js
```diff
@@ -8,7 +8,20 @@ NWC.model.AquaticBiologyFeaturesModel =  Backbone.Model.extend({
             sites : [],
             gages : [],
             hucs : [],
-            selected : []
-        }
+            selected : [],
+            pairs : []
+        },
+	associatePairs : function(s, g, action) {
+        this.updPairs = this.get('pairs').clone();
```
mbucknell: Need to fix the indentation. The object this.updPairs shouldn't go in the model's scope. You can declare as a local variable to this function.  It would be good to add tests for the associatePairs function.

nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -30,21 +36,15 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
...
-                    var selected = this.model.get('selected');
-                    var index = selected.indexOf(name);
-                        if (index > -1) {
-                            selected.splice(index, 1);
-                            this.model.set({ 'selected' : selected});        
-                        };
-			this.biodataGageMapView.unHighlightSite(name);
-                    }
-            } 
+                    $('#' + name + '-btn').prop('disabled', false);
```
mbucknell: Fix indentation.

nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -98,39 +98,54 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
...
 	    },
 	
 	    unHighlightGageRow : function(feature) {
-		var selectedLayer = feature.layer.name;
-		alert('Unselected ' + selectedLayer);
+		$('#gage-' + feature.attributes.STAID).removeClass('gage-selected');
+	    },
+    
+        removePair: function (evt) {
+            var $cb      = $(evt.target),
```
mbucknell: In general, I don't think you should put whitespace before the '=' sign.

nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -98,39 +98,54 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
...
+        onGageSelect : function(e){
                 e.preventDefault();
                 $cb = $(e.currentTarget);
                 gageID = $cb.attr('id');
                 this.router.navigate('/streamflow-stats/gage/' + gageID, {trigger : true});
-            },
+        },
 	    
-	    _displayMap : function() {	
+	    _displayMap : function(options) {
```
mbucknell: Fix indentation here and the next three functions

nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -98,39 +98,54 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
...
 			gageFeature : this.context.gages,
 			router : this.router,
 			el : $('#biodata-gage-selection-div'),
-			highlightGageRow : this.highlightGageRow,
-			unHighlightGageRow : this.unHighlightGageRow
+			model : this.model
 		});
             },
+	    
 	    highlightGageRow : function(feature) {
```
mbucknell: Where are highlightGageRow and unHighlightGageRow used ... I couldn't find any uses of them. If not used they should be removed.

kmschoep-usgs: I was going to leave them in until I got confirmation from Dave that he didn't want the gage rows highlighted.  If there won't be gage row highlighting I'll remove.

nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
```diff
@@ -98,39 +98,54 @@ NWC.view.AquaticBiologySelectFeaturesView = NWC.view.BaseView.extend({
...
 	
 	    unHighlightGageRow : function(feature) {
-		var selectedLayer = feature.layer.name;
-		alert('Unselected ' + selectedLayer);
+		$('#gage-' + feature.attributes.STAID).removeClass('gage-selected');
+	    },
+    
+        removePair: function (evt) {
+            var $cb      = $(evt.target),
+            siteId     = $cb.attr('id'),
```
mbucknell: It's a little weird to put this information in the button using the name and id attributes. A better way would be to use data-xxx attributes and then retrieve with the .data() jquery function. So the template would look like this for the button:
```
<button title="Remove Pair" data-siteId="{{site_id}}" data-gageId="{{gage_id}}" class="dismiss-btn">x</button>
```
and then use $cb.data('siteId') and $cb.data('gageId') to retrieve the values.

nwc/src/main/webapp/js/view/BiodataGageMapView.js
```diff
@@ -119,68 +138,80 @@ NWC.view.BiodataGageMapView = Backbone.View.extend({
...
+	    this.gageSelectControl.deactivate();
+	    this.model = options.model;
 	    Backbone.View.prototype.initialize.apply(this, arguments);
-	    this.render();
+	    this.render(options);
 	},
 
-	render : function() {
+	render : function(options) {
 		Backbone.View.prototype.render.apply(this, arguments);
```
mbucknell: This isn't really necessary because the default Backbone.View render is a noop.

nwc/src/main/webapp/templates/aquaticBiologySelectFeatures.html
```diff
@@ -34,7 +40,7 @@
                                     {{#each biodataSites}}
                                     <tr>
                                         <td class="xs-col-2">
-                                            <input type="checkbox" class="sites" id="{{attributes.SiteNumber}}" name="{{attributes.SiteNumber}}" value="{{attributes.SiteNumber}}"/>
+                                            <input type="checkbox" class="sites" id="ab-{{attributes.SiteNumber}}" name="{{attributes.SiteNumber}}" value="{{attributes.SiteNumber}}"/>
```
mbucknell: You might want to change this to use data-xxx attributes, rather than use id and name attributes to save data. It will make the code clearer.

mbucknell:
In general, the indentation needs to be fixed in many places. This may seem picky and it is, but it makes the code a little easier to read.  You might also look at adding tests for the displaying, adding, and removing pairs in the AquaticBiologySelectFeatureView.

Go ahead and merge if you want.

nwc/src/main/webapp/js/view/AquaticBiologySelectFeaturesView.js
kmschoep-usgs: thanks, I knew it was a hacky solution.  I need to read up more on jquery.

mbucknell: They are actually an HTML5 thing. Jquery has added convience functions to access them. See http://www.w3schools.com/tags/att_global_data.asp

PR #250, opened 2015-06-04
Watersmart 428

mamcderm-usgs:


nwc/src/main/webapp/js/view/DataDiscoveryView.js
```diff
@@ -37,19 +39,37 @@ NWC.view.DataDiscoveryView = NWC.view.BaseView.extend({
 	},
 
 	/**
+	 * This makes a Web service call to get data for display
+	 */
+	getData: function() {
```
mbucknell: This looks like it does the same thing as getProjects. Is there a reason to have both?

mbucknell: Does the getData have more detailed information than getProject? If so, maybe we can call getData and then just show/hide the details as necessary.

mamcderm-usgs: pushing a refactored verion...

nwc/src/main/webapp/js/view/DataDiscoveryView.js
```diff
@@ -37,19 +39,37 @@ NWC.view.DataDiscoveryView = NWC.view.BaseView.extend({
...
 			$(ev.currentTarget).html('-');
 			$(ev.currentTarget).prop('title', 'Hide details');
 			$.ajax({
-				url : CONFIG.endpoint.direct.sciencebase + '/catalog/item/' + projectId + '?format=json',
+				url : CONFIG.endpoint.direct.sciencebase + '/catalog/item/' + id + '?format=json',
 				dataType : "json",
 				success: function(data) {
-					$('#' + projectId).html(NWC.templates.getTemplate('projectDetail')({projectDetail : data}));
-					$('#' + projectId).show();
+					$('#' + id).html(NWC.templates.getTemplate('dataDiscoveryDetail')({detail : data}));
```
mbucknell: You should make a var  to hold $('#' + id) since it's reference multiple times in this function.

nwc/src/main/webapp/js/view/DataDiscoveryView.js
```diff
@@ -14,19 +19,19 @@ NWC.view.DataDiscoveryView = NWC.view.BaseView.extend({
...
+		$("#show-projects-button").prop('disabled', true);
+		this.getList(this.project);
 	},
 
 	/**
-	 * This makes a Web service call to get project list data for display
+	 * This makes a Web service call to get list data for display
 	 */
-	getProjectData: function() {
+	getList: function(type) {
```
mbucknell: Okay, I think I understand what's happening now. I like this refactor!

mbucknell:
:+1:

mamcderm-usgs:
@mbucknell - thank you for reviewing so quick and helpful comments.

PR #258, opened 2015-06-30
Watersmart 521

mamcderm-usgs:


mamcderm-usgs:
@mbucknell - if you have time, would you review this to see if ok to merge for Dave to take a look.  Alternatively, I can do a webex with Dave to review the UI if you think this needs a different approach.

nwc/src/main/webapp/js/view/BaseDiscoveryTabView.js
```diff
@@ -43,7 +43,8 @@ NWC.view = NWC.view || {};
 		},
 
 		events : {
-			'click .toggle-details-btn' : 'toggleDetails'
+			'click .toggle-details-btn' : 'toggleDetails',
+			'click a' : 'goToProject'
```
mbucknell: Can we make goToProject be something more generic? Perhaps goToPage?

nwc/src/main/webapp/js/view/BaseDiscoveryTabView.js
```diff
@@ -51,6 +52,7 @@ NWC.view = NWC.view || {};
 		 * @param {Object} options
 		 *     @prop {Jquery element} el - this is where the list will be rendered
 		 *     @prop {Boolean} showSummary (optional) - Whether the summary in the list is rendered.
+		 *     @prop {Boolean} showLink (optional) - Whether the link in the list is rendered.
```
mbucknell: Should add the router as a prop too.

nwc/src/main/webapp/js/view/ProjectView.js
```diff
@@ -0,0 +1,94 @@
...
+		this.projectDetailView.getDetails(options.projectId).done(function(data) {
+			$('#project-details').append('<h6>Title</h6>');
+			$('#project-details').append(data.title);
+			$('#project-details').append(NWC.templates.getTemplate('projectDetail')(data));
+		});
+
+		this.dataDetailView = new NWC.view.DataTabView({
+		});
+		
+		var _this = this;
```
mbucknell: We should be consistent. Elsewhere in the project we use self as the var name when assigning this to a variable.

nwc/src/main/webapp/js/view/ProjectView.js
```diff
@@ -0,0 +1,94 @@
...
+	/*
+	 * @constructor
+	 * @param {Object} options
+	 *     @prop {Jquery element} el - the html element where view is rendered
+	 */
+	initialize : function(options) {
+		// call superclass initialize to do default initialize
+		// (includes render)
+		NWC.view.BaseView.prototype.initialize.apply(this, arguments);
+		this.projectDetailView = new NWC.view.ProjectTabView({
```
mbucknell: This is my OCD nature, but I would rather have you set the showSummary and showLink options to false explictly here rather than rely on the defaults in the parent view. I'm also surprised that you don't have to specify el. How is this working?

mbucknell: Now that I've studied this further, I see what you are doing. We aren't really using the functionality in the ProjectTabView or DataTabView. Can we just reuse the templates rather than the whole view? You might be able to include the Title in the template by filtering it in/out using a context variable.

mbucknell:
@mamcderm-usgs, see the above comments. I'm hoping that we can just reuse the templates rather than ProjectTabView and DataDetailsTabView.

nwc/src/main/webapp/js/view/ProjectView.js
mamcderm-usgs: @mbucknell  - finally getting back to this.  I'm not sure what best to do here.  Unless I'm missing something, I am using the functionality in the ProjectTabView, specifically the getDetails() function, as well as, the properties.  Is there a way to use those without instantiating them and only use the template?  Otherwise, it seems like I would have to duplicate the function and properties in the ProjectView.  Given the way it is setup, I was just trying to avoid duplication but I agree it is a little "indirect".

mbucknell: If getDetails is really the only thing being reused, then I think it is ok to create a completely separate view which reuses the projectDetail template. It might be worthwhile to pull out into a url constants utility file, he detailsUrl function body so that it can be reused in both places. An alternative would be to have a a detailsView which would be used in both the TabView and this new view. That might be overkill though.  Call me if you want to talk further.

mamcderm-usgs:
@mbucknell - would you take another look when you get a chance.  I've simplified to not use the tab views and only use the templates.  It could probably still be optimized some but wanted to make sure this is the right idea.

nwc/src/main/webapp/js/view/ProjectView.js
```diff
@@ -0,0 +1,114 @@
...
+
+	events: {
+		'click #back-button': "back",	
+	},
+
+	detailsUrl : function(id) {
+		return CONFIG.endpoint.direct.sciencebase + '/catalog/item/' + id + '?format=json';
+	},
+
+	listUrl : function(id) {
```
mbucknell: Don't think you need this at all.

nwc/src/main/webapp/templates/project.html
```diff
@@ -0,0 +1,15 @@
+<div class="row" style="margin: 0px;">
+	<div class="col-xs-12 nwc-container">
+		<div class="row">
+			<div class="col-xs-12">
+				<h3>Data Discovery - Project Detail Page</h3>
+			</div>
+		</div>
+		<div class="row" style="margin: 0px;"><button id="back-button" class="btn btn-primary btn-xs">Back</button></div>
```
mbucknell: Styles should go in the style sheet.

nwc/src/main/webapp/js/view/ProjectView.js
```diff
@@ -0,0 +1,114 @@
...
+		this.getDataList(options.projectId).done(function(dataList) {
+			if (dataList.items.length == 0) {
+				$('#data-details').append('<div class="row" style="margin: 0px;">None available</div>');				
+			}
+			else{
+				var i;
+				for (i=0; i < dataList.items.length; i++) {
+					self.getDetails(dataList.items[i].id).done(function(data) {
+						$('#data-details').append(NWC.templates.getTemplate('dataDetail')(data));
+						$('#data-details').append('<div class="row" style="margin: 0px;border-bottom: 1px solid black;"></div>');
```
mbucknell: Is this being used for spacing? A better option would be to style the #data-details div to include a bigger margin and put it in the style sheet.

nwc/src/main/webapp/js/view/ProjectView.js
```diff
@@ -0,0 +1,114 @@
...
+			if (dataList.items.length == 0) {
+				$('#data-details').append('<div class="row" style="margin: 0px;">None available</div>');				
+			}
+			else{
+				var i;
+				for (i=0; i < dataList.items.length; i++) {
+					self.getDetails(dataList.items[i].id).done(function(data) {
+						$('#data-details').append(NWC.templates.getTemplate('dataDetail')(data));
+						$('#data-details').append('<div class="row" style="margin: 0px;border-bottom: 1px solid black;"></div>');						
+					});
```
mbucknell: This is what I had in mind. Again, you might consider putting the url generation in some kind of utilities package.  Otherwise besides the comments about styling, this looks good.

PR #259, opened 2015-07-09
Watersmart 521

mamcderm-usgs:


mamcderm-usgs:
@mbucknell - These changes were based on these comments from Dave in:
https://internal.cida.usgs.gov/jira/browse/WATERSMART-524

Not complex changes but wanted to check with you before merging in case you had any concerns about the user interface impact.

mbucknell:
My only comment is to not leave in commented out code. Otherwise its fine. :+1:

PR #261, opened 2015-07-13
Watersmart 524

mamcderm-usgs:
@mbucknell - This seems to work but let me know if you see any better practices I should use.  It also seems like the tab views could be refactored into three distinct views rather than being "sub" views of the DataDiscovery view.  That way each list would be refreshed when clicking on the tab, and possibly simplify the design.  However, for this iteration I took the approach of minimal change to engage the backbone history.

### DIFF
--- a/nwc/src/main/webapp/js/controller/NWCRouter.js
+++ b/nwc/src/main/webapp/js/controller/NWCRouter.js
@@ -26,8 +26,10 @@ NWC.controller.NWCRouter = Backbone.Router.extend({
 		'streamflow-stats/model-info' : 'streamflowStatsModeledInfo',
 		'aquatic-biology' : 'aquaticBiology',
 		'aquatic-biology/select-features' : 'aquaticBiologySelectFeatures',
-		'data-discovery' : 'dataDiscovery',
-		'data-discovery/project/:projectId' : 'dataDiscoveryProject'
+		'data-discovery/show-projects' : 'dataDiscoveryProjects',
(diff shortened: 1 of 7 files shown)


nwc/src/main/webapp/js/controller/NWCRouter.js
```diff
@@ -95,11 +97,19 @@ NWC.controller.NWCRouter = Backbone.Router.extend({
             });
 	},
 
-	dataDiscovery : function() {
-		this.showView(NWC.view.DataDiscoveryView);
+	dataDiscoveryProjects : function() {
```
mbucknell: Rather than repeating dataDiscoverProjects, dataDiscoveryData, and dataDiscoveryPublications, you can have one function that is parameterized (like the projectDetail is with projectId). So you would parameterize it with dataKind which would be one of show-projects, show-publications, or shoe-data.

mbucknell:
@mamcderm-usgs, looks good. I would not create three separate child views for the three different dataDiscovery view. I don't think refreshing the data is not necessary since the data retrieved is not continually changing. I think most of the complexity is actually in the tab views.

nwc/src/main/webapp/js/controller/NWCRouter.js
mamcderm-usgs: Sounds good.  I had it that way but switched it when I was having an issue, even though I assume there is no difference between a parameter and a "path".  I'll switch it back.

PR #266, opened 2015-07-22
Watersmart 526

mamcderm-usgs:
This is first step of implementing this feature.  Just getting side by side statistics working.


nwc/src/main/webapp/js/view/BaseStreamflowStatsDataView.js
```diff
@@ -5,9 +5,12 @@ NWC.view = NWC.view || {};
 NWC.view.BaseStreamflowStatsDataView = NWC.view.BaseView.extend({
 
 	events : {
-		'click #calculate-stats-button' : 'calculateStats',
-		'click #available-statistics input' : 'calculateStatsEnable',
-		'click #download-stats-button' : 'downloadStats'
+		'click #calculate-stats-button-left' : 'calculateStats',
+		'click #calculate-stats-button-right' : 'calculateStats',
+		'click #available-statistics-left input' : 'calculateStatsEnableLeft',
+		'click #available-statistics-right input' : 'calculateStatsEnableRight',
```
mbucknell: Rather than having a calculateStatsEnableLeft and calculateStatsEnableRight and using the {{pane}} template variable to create unique ids, you could make these class names and use selectors like .parent() to get to the correct available-statistics input to disable. You can use the event object itself enable/disable the calculate button. @thongsav-usgs and I have had discussions about whether we should use class names rather than ids in our selectors in Backbone views and I'm coming around to his thinking on this. The reasoning is that it should make the code more reusable/portable. Also using this.$el.find where appropriate to select DOM elements is a good practice (this is just FYI).

mbucknell:
You might consider creating a childView for the calculate stats portion of the streamflow page. What I'm thinking is a childView that handles only one set of calculations so for the parent views, you would create two childViews. 

Don't feel that have to do this ... it's only something that to consider. I haven't thought it through. I would like to see class names used rather than ids for DOM element selection.

nwc/src/main/webapp/js/view/BaseStreamflowStatsDataView.js
mamcderm-usgs: I tried going down that path but I couldn't get it to work.  It's likely I just didn't completely understand the dom well enough, but the problem with this seemed to be that the button being enabled/disabled was not a parent of the input selector.  Anyway, let me know if you prefer I rework this section...assuming I may need some guidance.

mamcderm-usgs:
@mbucknell - I've pushed up an alternative solution based on our discussion.  It's somewhat hasty in that I'm trying to get something that could be mergable before I go on leave for a few days but it does work locally.   It should be cleaned up as well except I left the partial template for now.  If you have time to review today, please let me know what if anything should be done to make it mergable.  Thanks.

nwc/src/main/webapp/js/view/BaseStreamflowStatsDataView.js
```diff
@@ -10,18 +10,25 @@ NWC.view.BaseStreamflowStatsDataView = NWC.view.BaseView.extend({
...
-		}
-		this.context.streamStatsOptions = NWC.dictionary.statGroups;
 
 		NWC.view.BaseView.prototype.initialize.apply(this, arguments);
-	},
 
+		this.CalculateStatsView = new NWC.view.StreamflowStatsCalculateStatsView({
+			el : $('.calculate-stats'),
+			years : this.context.years || null
+		});
```
mbucknell: I'm surprised this works. So you only have to create a single StreamflowStatsCalculateStatsView? This then creates both the left and the right?

mbucknell:
If it's working (and the tests run), I think it's okay to merge. There's a lot that will need to be cleaned up. If you wait though, I'm hoping the merge won't be too bad. But I understand the desire not to have to deal with it!

If you do merge, I'll do a better review on this tomorrow as I need to leave very soon.

nwc/src/main/webapp/js/view/BaseStreamflowStatsDataView.js
mamcderm-usgs: Yes, it appears to work ok that way.  I originally had two then I thought I'd try it.  Based on how the class selector works, this does seem to make sense.

mamcderm-usgs:
I'm running out of time and as usual I've forgotten to run the tests.  So, I think I'll wait until next week to get this to a better state and merge.  Thanks.

mamcderm-usgs:
@mbucknell - as you expected, everything fell into place when I explicitly instantiated two views.  Please take a look when you get a chance to see if it can be merged.  Hopefully, it is close.  I haven't started to try and toggle between one pane and two but will work on that if this design is sound.

nwc/src/main/webapp/js/view/StreamflowCalculateStatsView.js
```diff
@@ -0,0 +1,126 @@
...
+			'click #available-statistics input' : 'calculateStatsEnable',
+			'click #download-stats-button' : 'downloadStats'
+		},
+		
+		/*
+		 * @constructs
+		 * @param {Object} options
+		 *     @prop {Jquery element} el - this is where the list will be rendered
+		 *     @prop {object} streamStatsOptions - this is the list of stat types
+		 *     @prop {object} years (optional) - when using the streamflowHucData view
```
mbucknell: A couple of things. Comments should reflect how this unit is used, not how the instantiater of the view will use the option so you should probably remove references to streamflowHucData view and streamflowGageData view.

We should also figure out why the gage is not using the years. There are year pickers on the page.

nwc/src/main/webapp/js/view/StreamflowCalculateStatsView.js
```diff
@@ -0,0 +1,126 @@
...
+		 *     @prop {object} gageId (optional) - when using the streamflowGageData view
+		 *     @prop {object} hucId (optional) - when using the streamflowHucData view
```
mbucknell: Does the view use gageId or hucId? The only place I saw was in the context and didn't see any references to either in the template. I think they could be removed.

nwc/src/main/webapp/js/view/StreamflowCalculateStatsView.js
```diff
@@ -0,0 +1,126 @@
...
+		 *     @prop {object} getStats - function
+		 *     @prop {object} getStatsTsvHeader - function
+		 *     @prop {object} getStatsFilename - function
```
mbucknell: Consider adding a little more detail to the description of these function properties.

nwc/src/main/webapp/js/view/StreamflowStatsHucDataView.js
```diff
@@ -71,13 +71,30 @@ NWC.view.StreamflowStatsHucDataView = NWC.view.BaseStreamflowStatsDataView.exten
 		});
 		this.map.addLayer(this.hucLayer);
 
-		$.extend(this.events, NWC.view.BaseStreamflowStatsDataView.prototype.events);
-		NWC.view.BaseStreamflowStatsDataView.prototype.initialize.apply(this, arguments);
+		NWC.view.BaseView.prototype.initialize.apply(this, arguments);
 		this.map.zoomToExtent(this.map.getMaxExtent());
 
+		this.CalculateStatsViewLeft = new NWC.view.StreamflowCalculateStatsView({
```
mbucknell: OCD comment: should probably be this.calculateStatsViewLeft (lowercase first letter).

mbucknell:
Other than my confusion around whether years should be passed in for the gage view, it looks good. :+1:

nwc/src/main/webapp/js/view/StreamflowCalculateStatsView.js
mamcderm-usgs: They are used in the "passed in" functions.   I'm guessing there is another way to reference them that would work without having to pass to sub view?

nwc/src/main/webapp/js/view/StreamflowCalculateStatsView.js
mamcderm-usgs: The "years" thing was strange.  The GageDataView and HucDataView had it implemented differently and I didn't stop to analyze why...I generally assume there was a good reason.  I would assume they could be done the same way.  Should I go ahead and refactor?

nwc/src/main/webapp/js/view/StreamflowCalculateStatsView.js
mbucknell: This is where you would bind the "this" to the getStatsTsvHeader, etc, so that it can access its own scope, rather than relying on these values to be passed in to the sub view. I do a similar thing when creating the this.streamflowPlotView for the getDataSeriesPromise function.

nwc/src/main/webapp/js/view/StreamflowCalculateStatsView.js
mbucknell: When I look at the existing code, it looks like it is used ... so I must be missing something. But if there is a data picker we should be paying attention to it, I think. Or it shouldn't be there for the calculate stats for the gage view. Might ask @dblodgett-usgs .

mbucknell: In the app out on prod, the gage page does use the dates when calculating statistics

nwc/src/main/webapp/js/view/StreamflowCalculateStatsView.js
mamcderm-usgs: awesome!  thank you

nwc/src/main/webapp/js/view/StreamflowCalculateStatsView.js
mamcderm-usgs: Yes, both views use the selected dates for calculating statistics.  For some reason, the GageDataView populates the date picker in the view whereas the HucDataView populates the date picker in the template.  Nothing obvious to me for why it was done that way....

mbucknell: Was one dynamic? If memory serves, the Gage data view's date ranges are retrieved from a service call. Maybe the Huc Data View's date ranges are static?

mamcderm-usgs: Yes, I believe that's true.  I'm going to try and refactor so that both are rendered in the template.

mbucknell: Just be aware that that may delay rendering of the template. That may be okay since you won't be able to use the section until the date pickers are set up anyway.

PR #268, opened 2015-07-27
Make dev onboarding easier

cschroed-usgs:


cschroed-usgs:
Updating documentation of necessary configuration to abide by new convention that @isuftin @mbucknell and I discussed.

README.md
```diff
@@ -3,3 +3,30 @@
...
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<Context>
+    <!-- Default set of monitored resources -->
+    <WatchedResource>WEB-INF/web.xml</WatchedResource>
+
+    <!-- disable session persistence across Tomcat restarts -->
+    <Manager pathname="" />
+
+	<Environment name="nwc.endpoint.geoserver" value="http://cida-eros-wsdev.er.usgs.gov:8081/geoserver/" type="java.lang.String" override="true"/>
```
mbucknell: @cschroed-usgs, Since this is a public repo, do you think we should expose our dev resources here?

cschroed-usgs: I thought about that. First, these values were already in the repo in the exampleContext.xml, so the changes I merged are not letting any cats out of the bag. Second, Ben Feinstein ok'd us putting dev urls on github in the form of pull-request-specific Jenkins build results. The potential risk for these dev resources is equivalent to the risk for the Jenkins urls. Since the Jenkins urls are considered an acceptable risk, I figured these were too.

There's another non-security concern: external collaborators. This example context is not useful to them because they don't have access to our network. Should we put public-facing endpoints here instead? Who should the default values cater to? Developers or collaborators? The defaults for one are not super useful for the other.

mbucknell: In this case, we have the actual values in STASH for our configuration deployer. Since developers already have access to the values, I think it makes more sense to make this useful to our external collaborators.

thongsav-usgs: I personally am not super concerned about the network infrastructure being leaked. I do prefer NOT checking in those URLs though, as I find they are more red herrings when network infrastructure changes.

I propose some kind of descriptive tag in place of the URLs (even if the JNDI name is self explanatory). EG value="[URL pointing to geoserver with X data]"

cschroed-usgs: @mbucknell Cool, I didn't know that we had an automated config repo for NWC. w00t!  Hopefully more projects start going that way.

@thongsav-usgs I think your proposal makes sense if the example context caters to devs. Do you think that it's still applicable if devs already have access to valid config values via that stash repo? I don't think that our public production endpoints for apps will change much.

cschroed-usgs: On the other hand, if we already have the config as runnable documentation in stash, should we bother duplicating (and forgetting to update it) here?

cschroed-usgs: If that wasn't clear, what I'm proposing is that we either leave this example context here with production urls, or strip it out entirely. Thoughts?

thongsav-usgs: In the case of Readme.md, I can live with an actual URL, but am still not a fan. A lot of apps default to hardcoded URLs, and I feel like this is in the same boat.

I do not like any URLs in source, because it works until it doesn't. I prefer the explicit step of users having to consciously configure.

thongsav-usgs: On second thought, for things like the txpub URL, you either know it or you don't. And if you don't you probably have a slim chance of getting the app to work.

mbucknell:
If our goal is to have outside the USGS folks by able to run the project and we are specifying urls, then it seems to me we should use urls that are available outside of our network.

README.md
cschroed-usgs: Yeah, some of them are definitely so specific that it's probably better to put them in than to describe what kind of service interface the endpoint should provide. Do you favor a mix of hardcoded prod urls for those super-specific values and instructions for users for the more generic ones (geoserver, thredds, etc), or are you in favor of hardcoding em all with prod urls?

I share your concern about the example context going out of date. I'm not sure which is better it looks like we've got:
 1. giving people who fork our repos a potentially broken example context with all values filled in with prod urls
 2. giving people templates contexts with instrucitons on how to fill in missing values (maybe some super-specific urls predefined)
 3. not providing an example context at all, helping every collaborator individually

Since this project, and hopefully others are moving to checking configs into private repos, we're providing example configs for collaborators who might want to fork our apps. This doesn't happen often. Kind of a chicken-and-egg thing -- are people not forking the repos because they aren't interested, because our apps are too specific to be re-usable, or because they are intimidated by the inadequacy of our instructions and example configs? This is kind of a marketing thing.

If our goal is to incentivize forking and running, making the initial apparent barrier to entry low is important. So is reducing the back-and-forth between CIDA devs and collaborators.

Option 1 has the lowest apparent barrier to entry because in the best case collaborators have to do no work to get it to run, but it is more likely that collaborators will contact us because configs are out of date than Option 2. Collaborators would fork, locally clone, configure the app, and get to the point of running it locally before they ran into problems. The appearance of ready-to-go configs would convince collaborators to get kind of invested before they ran into problems.
Option 2 has the middle apparent barrier to entry because collaborators must do some work for it to run in the best case, and the middle risk that collaborators will need to contact us to run the app.
Option 3 has the highest apparent barrier to entry, and it is the most likely to require that collaborators will contact us to run the app, but the number is conditioned on the barrier to entry, so the contact risk is somewhat deflated. In other words, 100% of the people that try to run the project would need to contact us to get it to work, but that number would be smaller because the lack of instructions would discourage people from trying to run it all.

PR #272, opened 2015-07-28
Watersmart 528

mbucknell:
A lot of this is actually refactoring the WaterBudgetHucDataView into using sub views, the new WaterBudgetPlotView and CountyWaterUseView. The WaterBudgetCountyHucDataView was eliminated and instead the huc/county view is created via the WaterBudgetHucDataView with much of the old functionality move to the new sub view CountyWaterUseView.

I'm also trying to use this.$el.find when selecting DOM elements in the views that I have touched.


nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -14,77 +14,122 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
...
 		NWC.view.BaseView.prototype.initialize.apply(this, arguments);
 
 		this.setUpHucPlotModel();
 
-		this.buildHucMap(this.hucId);
-		this.getHucData(this.hucId);
+		// Create additional sub views as needed
+		$plotContainer = this.$el.find('#huc-plot-container');
+		if (this.compareHucId) {
+			$plotContainer.html(NWC.templates.getTemplate('hucComparePlotViewContainer')());
```
mbucknell: I had this html hard coded but decided to move it to a template. Now I'm not sure if its worth it.

mbucknell:
@thongsav-usgs , could you review if you have time? This is way larger than the last, but is mostly refactoring to use subviews.

nwc/src/main/webapp/js/init.js
```diff
@@ -9,8 +9,10 @@ $(document).ready(function() {
 	var TEMPLATES = [
 		'home',
 		'waterbudget',
+		'waterbudgetPlot',
+		'hucComparePlotViewContainer',
 		'waterbudgetHucData',
-		'waterbudgetCountyData',
+		'countyWaterUse',
```
thongsav-usgs: Do you load and cache these somewhere? AQCU right now seems chatty because we reload templates each time the a view is initialized (and re-initialized over and over).

nwc/src/main/webapp/js/view/CountyWaterUseView.js
```diff
@@ -0,0 +1,245 @@
...
+			};
+
+			// Create inset map containing the county and the huc
+			baseLayer = NWC.util.mapUtils.createWorldStreetMapLayer();
+			this.map = NWC.util.mapUtils.createMap([baseLayer], [new OpenLayers.Control.Zoom(), new OpenLayers.Control.Navigation()]);
+
+			this.countyLayer = NWC.util.mapUtils.createCountyFeatureLayer(this.fips);
+			this.countyLayer.events.on({
+				featureadded: function(event){
+					this.countyName = event.feature.attributes.full_name.capitalize(true);
```
thongsav-usgs: Do these handlers deserve to be promoted to view functions? I can imagine wanting to invoke this functionality (minus the event parsing) from else where.

nwc/src/main/webapp/js/view/CountyWaterUseView.js
```diff
@@ -0,0 +1,245 @@
...
+			this.hucLayer = NWC.util.mapUtils.createHucFeatureLayer([this.hucId]);
+			this.hucLayer.events.on({
+				loadend : function() {
+					hucLayerLoaded.resolve();
+				}
+			});
+			this.map.addLayer(this.hucLayer);
+
+			// Set up the code to render the water use plot once the county and huc layers have been loaded
+			$.when(countyLayerLoaded, hucLayerLoaded).done(function() {
```
thongsav-usgs: Same question... I might be falling out of love with anonymous functions...

nwc/src/main/webapp/js/view/CountyWaterUseView.js
```diff
@@ -0,0 +1,245 @@
...
+			});
+
+			return deferred.promise();
+		},
+
+		/**
+		 * Update the water use chart using the countyPlotModel and waterUseDataSeries properties
+		 */
+		chartWaterUse : function() {
+			var chartDivEl = this.$el.find('#waterUsageChart');
```
thongsav-usgs: I should verify, but I simply use this.$('selector') and I'm *pretty sure* it does the same thing and limits the the extent to which jQuery looks to just what belongs to the backbone view.

nwc/src/main/webapp/js/view/CountyWaterUseView.js
```diff
@@ -0,0 +1,245 @@
...
+
+			NWC.util.WaterUsageChart.setChart(chartDivEl, chartLegendDivEl, values, labels, ylabel,
+				NWC.util.Units[measurementSystem][plotNormalization].precision);
+			return;
+		},
+
+		setUpCountyPlotModel : function() {
+			// add listeners to model
+			this.countyPlotModel = new NWC.model.WaterBudgetCountyPlotModel();
+			this.listenTo(this.countyPlotModel, 'change:units', this.updateCountyUnits);
```
thongsav-usgs: Shoot, I'm not sure why I stopped using listenTo... we've got some ugly patterns developing in AQCU where we are duplicating values in multiple models (parent/sub view models) just to get model event triggering. @dpattermann-usgs

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -14,77 +14,122 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
 
 	templateName : 'waterbudgetHucData',
 
-	ETA : "eta",
-	DAY_MET : "dayMet",
-
 	events: {
 		'click #counties-button' : 'displayCountyMap',
```
thongsav-usgs: So. Many. Ids. :smiling_imp:

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -200,46 +175,24 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
...
-		return filename;
+		this.setButtonActive(this.$el.find('#daily-button'), newTimeScale === 'daily');
+		this.setButtonActive(this.$el.find('#monthly-button'), newTimeScale === 'monthly');
 	},
 
 	remove : function() {
 		if (Object.has(this, 'hucCountyMapView')) {
 			this.hucCountyMapView.remove();
 		}
+		this.plotView.remove();
```
thongsav-usgs: Ugh, another point we're probably overlooking in AQCU. Need to review our remove functions @dpattermann-usgs

thongsav-usgs:
I will never ever stop whining about using dom ids (unless browsers change how they are treated), but aside from that :+1: :smile:

nwc/src/main/webapp/js/init.js
mbucknell: There are initialized when the app is loaded: https://github.com/USGS-CIDA/nwc/blob/master/nwc/src/main/webapp/js/init.js

nwc/src/main/webapp/js/view/CountyWaterUseView.js
mbucknell: I prefer being explicit and it seems like using this.$ rather than this.$el.find is using something that is implicit and to my knowledge not documented.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
mbucknell: I have come around to your way of thinking. And I'm using classes where I can, but this was existing code and I already made a lot of other changes.

mbucknell:
I'm not sure I understand what you mean about how browser treat dom ids. But I agree that views should use classes. An exception to this is that some libraries, like OpenLayers, require an id, for instance when creating the map. I am not being consistent here. Sometimes I hard code it, sometimes I pass it in to the view. Since the templates are so tightly coupled to a view I should probably hard code it as a view level constant so it can possibly by overridden if the view is extended. 

I'm starting a list of things to clean up in NWC. I'll add to the list using classes rather than ids in the views, but won't fix for now.

nwc/src/main/webapp/js/view/CountyWaterUseView.js
mbucknell: My bad, this is a documented feature. I'll start using it but probably won't change these now.

PR #273, opened 2015-08-03
Watersmart 528

mbucknell:
Created a new subView, HucInsetMapView, and used it to create the water budget huc data page huc inset map. If we are comparing hucs then two inset maps are shown.

### DIFF
--- a/nwc/src/main/webapp/js/utils/templateLoader.js
+++ b/nwc/src/main/webapp/js/utils/templateLoader.js
@@ -29,6 +29,7 @@ NWC.util.templateLoader = function() {
 			templates[names[i]] = '';
 			loadingDeferreds.push($.ajax({
 				url : 'templates/' + names[i] + '.html',
+				cache : false,
 				success : function(data) {
 					templates[this] = Handlebars.compile(data);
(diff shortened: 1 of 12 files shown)


nwc/src/main/webapp/js/utils/templateLoader.js
```diff
@@ -29,6 +29,7 @@ NWC.util.templateLoader = function() {
 			templates[names[i]] = '';
 			loadingDeferreds.push($.ajax({
 				url : 'templates/' + names[i] + '.html',
+				cache : false,
```
mbucknell: This is an unrelated change, but decided to set cache to false when loading the templates so that they are always loaded and we shouldn't have to clear the browser cache to read new templates.

mbucknell:
@mamcderm-usgs, here is another example of using sub view, which I'm now using to create the huc inset map. On interesting thing in this view is using hucFeatureLoadedPromise to  allow code outside the view to do something (in this case enable some buttons) when the huc feature has been loaded.

mamcderm-usgs:
:+1: Looks good to me.

PR #274, opened 2015-08-04
Watersmart 518

cschroed-usgs:
This pull request prefixes the client-side routes with a hashbang so that Google and Bing's ajax-compatible crawlers will gain visibility into the app. Some server-side stuff needs to happen too. I'll put that in a separate pull request.


mbucknell:
Seems pretty straight forward. Looking forward to seeing what changes need to happen on server side.
:+1: 

@cschroed-usgs, I'll let you merge when you are ready. When you do, please add a comment to the NWC slack chat channel to warn people about the changes to the urls.

PR #275, opened 2015-08-06
watersmart-527; add plot view to streamflow view and plug missing val…

mamcderm-usgs:
…ues in plot

### DIFF
--- a/nwc/src/main/webapp/js/view/StreamflowStatsGageDataView.js
+++ b/nwc/src/main/webapp/js/view/StreamflowStatsGageDataView.js
@@ -13,7 +13,7 @@ NWC.view.StreamflowStatsGageDataView = NWC.view.BaseView.extend({
 	events : {
 		'click .show-plot-btn' : 'plotStreamFlowData'
 	},
-
+ 
 	/*
 	 * Query NWIS for information about this.context.gageId. If the call fails
 	 * the deferred will be rejected with a default start and end date. Otherwise the start and
@@ -202,7 +202,11 @@ NWC.view.StreamflowStatsGageDataView = NWC.view.BaseView.extend({
 			// Enable show plot button
 			$('.show-plot-btn').removeProp('disabled');
 		});
+
+		this.dataSeriesLoaded = $.Deferred();
+
 		$.when(nwisDataRetrieved, featureLoaded).done(function () {
+			self.getDataSeries();
 			$('#loading-indicator').hide();
 		});
 
@@ -234,11 +238,18 @@ NWC.view.StreamflowStatsGageDataView = NWC.view.BaseView.extend({
 	},
 
 	/*
+	 * @returns Jquery promise which is resolved if getDataSeries() returns successfully in init.
+	 */
+	getDataSeriesPromise : function() {
+		return this.dataSeriesLoaded.promise();
+	},
+
+	/*
 	 * @returns Jquery promise which is resolved with the data series if it is successfully retrieved. If
 	 * unsuccessful is is rejected and forwards on the text response of the bad request
 	 */
-	getDataSeriesPromise : function() {
-		var deferred = $.Deferred();
+	getDataSeries : function() {
+		var self = this;
 
 		var startDate = this.dates.startDate;
 		var endDate = this.dates.endDate;
@@ -266,7 +277,7 @@ NWC.view.StreamflowStatsGageDataView = NWC.view.BaseView.extend({
 				});
 
 				if (dataTable.length === 0) {
-					deferred.reject('No data available to plot');
+					self.dataSeriesLoaded.reject('No data available to plot');
 				}
 				else {
 					dataSeries.data = dataTable;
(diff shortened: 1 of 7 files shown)


nwc/src/main/webapp/js/view/StreamflowStatsGageDataView.js
```diff
@@ -102,8 +102,12 @@ NWC.view.StreamflowStatsGageDataView = NWC.view.BaseView.extend({
...
 		NWC.view.BaseView.prototype.render.apply(this, arguments);
 		this.map.render(this.insetMapDiv);
-		this.streamflowPlotView = new NWC.view.StreamflowPlotView({
-			el : this.$el.find('.streamflow-plot-container'),
+		this.streamflowPlotViewLeft = new NWC.view.StreamflowPlotView({
+			el : this.$el.find('#left-plot'),
+			getDataSeriesPromise : this.getDataSeriesPromise.bind(this)
+		});
+		this.streamflowPlotViewRight = new NWC.view.StreamflowPlotView({
+			el : this.$el.find('#right-plot'),
```
mbucknell: This will retrieve the same data again. There's probably a way to do this so that it's only retrieved once. If you have time you might consider how to do this but this works for now.

nwc/src/main/webapp/js/view/StreamflowStatsGageDataView.js
```diff
@@ -286,7 +293,13 @@ NWC.view.StreamflowStatsGageDataView = NWC.view.BaseView.extend({
...
 		ev.preventDefault();
 
-		this.streamflowPlotView.plotStreamflowData(plotTitle).done(function(dataSeries) {
+		this.streamflowPlotViewLeft.plotStreamflowData(plotTitle).done(function(dataSeries) {
+			self.dataSeries = dataSeries;
+			self.$el.find('.show-plot-btn').hide();
+		}).fail(function(args) {
+			alert('Retrieving data for this plot failed with error: ' + args[0]);
+		});
+		this.streamflowPlotViewRight.plotStreamflowData(plotTitle).done(function(dataSeries) {
```
mbucknell: If you keep it so the data is retrieved twice, I would assign the dataSeries when both promise have been resolved. I think it makes sense to hide the show plot btn before the call to plotStreamflowData. We don't want the user clicking it twice.

nwc/src/main/webapp/js/view/StreamflowStatsHucDataView.js
```diff
@@ -177,7 +181,14 @@ NWC.view.StreamflowStatsHucDataView = NWC.view.BaseView.extend({
 		var plotTitle = 'Modeled Streamflow for the ' + this.hucName + ' Watershed.';
 		ev.preventDefault();
 
-		this.streamflowPlotView.plotStreamflowData(plotTitle).done(function(dataSeries) {
+		this.streamflowPlotViewLeft.plotStreamflowData(plotTitle).done(function(dataSeries) {
```
mbucknell: Same comments as with the gage view.

mbucknell:
:+1:  Merge when you want.

nwc/src/main/webapp/js/view/StreamflowStatsGageDataView.js
mamcderm-usgs: Any thoughts on how to best do this?  Get the promise first and then pass it in, rather than pass in the function?  Or, maybe check in getDataSeriesPromise() if a promise exists already?

mbucknell: The StreamflowPlotView expects a function that returns a promise. You could declare the deferred that is now in getDataSeriesPromise as a view level property (I would create it in the initialize). Then create the two streamflowPlotView's with a function that returns this deferred's promise. In plotStreamFlowData, you can then make the call getDataSeriesPromise (basically what is now getDataSeriesPromise without the deferred creation). This should do the trick.

nwc/src/main/webapp/js/view/StreamflowStatsGageDataView.js
mamcderm-usgs: @mbucknell - I'm get lost on following the deferred's.  Can you explain the purpose of assigning the dataSeries in line 297?  It doesn't seem necessary to me so I must be missing some connection.  Thanks.

mbucknell: In this case it isn't necessary or used. In the StreamflowStatsHucDataView, the dataSeries property is used to save the data for download. It could be removed here.

mamcderm-usgs: @mbucknell - in order to assign the dataSeries after both promises have been resolved, were you thinking to do that by chaining with "then", or is there a better way to do that?

mbucknell: I would use "when". Here's an example: https://github.com/USGS-CIDA/nwc/blob/master/nwc/src/main/webapp/js/init.js.

mamcderm-usgs:
@mbucknell - I've modified the code to only pull plot data once.  Seems to work but please check to see if I've introduced any unintended effects.

nwc/src/main/webapp/js/view/StreamflowStatsGageDataView.js
```diff
@@ -198,7 +202,11 @@ NWC.view.StreamflowStatsGageDataView = NWC.view.BaseView.extend({
 			// Enable show plot button
 			$('.show-plot-btn').removeProp('disabled');
 		});
+
+		this.dataSeriesLoaded = $.Deferred();
+
 		$.when(nwisDataRetrieved, featureLoaded).done(function () {
+			self.getDataSeries();
```
mbucknell: It would be better to wait to make this call to getDataSeries() until the user has requested the plot in plotStreamFlowData. Otherwise this looks good.

PR #280, opened 2015-08-13
Watersmart 539

mamcderm-usgs:


nwc/src/main/webapp/js/view/StreamflowStatsMapView.js
```diff
@@ -53,74 +52,67 @@ NWC.view.StreamflowStatsMapView = NWC.view.BaseSelectMapView.extend({
...
+		this.hucLayer.addOptions({attribution: '<img src="' + this.legendUrl(this.hucLayer.params.LAYERS, "") + '"/>'});
+		
 		this.legendControl = new OpenLayers.Control.Attribution();
-
-		var gageFeatureInfoHandler = function(responseObject) {
-			if (responseObject.features.length > 1) {
+		
+		var featureInfoHandler = function(responseObject) {
+			if (responseObject.features.length > 2) {
 				this.showWarningDialog('Multiple gages were selected. Please zoom in and select a single gage.')
```
mbucknell: The text should change to say 'Multiple features were selected. Please zoom in and select a single feature.'

nwc/src/main/webapp/js/view/StreamflowStatsMapView.js
```diff
@@ -53,74 +52,67 @@ NWC.view.StreamflowStatsMapView = NWC.view.BaseSelectMapView.extend({
...
 			}
-			else if (responseObject.features.length === 1) {
+			else if (responseObject.features.length === 2) {
 				this.router.navigate('#!streamflow-stats/gage/' + responseObject.features[0].attributes.STAID, {trigger : true});
```
mbucknell: I don't think you should assume the first feature in the array is a gage. There are cases where two huc features could be selected. You still need to make sure that it is a gage before proceeding.

mbucknell: For that matter you could have also selected two gages, in which case you should show the warning dialog. So in the case of two features returned, you have to check if one is a gage and one is a huc before proceeding to the gage route. Otherwise, you'll want to show the warning dialog.

nwc/src/main/webapp/js/view/StreamflowStatsMapView.js
```diff
@@ -53,74 +52,67 @@ NWC.view.StreamflowStatsMapView = NWC.view.BaseSelectMapView.extend({
...
+				if (responseObject.features[0].fid.has('gages')) {
+					this.router.navigate('#!streamflow-stats/gage/' + responseObject.features[0].attributes.STAID, {trigger : true});					
+				}
+				else {
+					var sortedFeature;
+					if (responseObject.features.length > 0) {
+						sortedFeature = responseObject.features.min(function(el) {
+							var result = Infinity;
+
+		                    var mi2 = parseFloat(el.data.mi2);
```
mbucknell: Need to fix indentation.

nwc/src/main/webapp/js/view/StreamflowStatsMapView.js
```diff
@@ -53,74 +52,67 @@ NWC.view.StreamflowStatsMapView = NWC.view.BaseSelectMapView.extend({
...
-					this.showWarningDialog("Hydrologic model results are not valid for watersheds this large (" + km2.round(0) + " km<sup>2</sup>). Try looking for a nearby observed flow gage.");
-				} else {
-					this.router.navigate('#!streamflow-stats/huc/' + sortedFeature.attributes.huc12, {trigger : true});
+			else if (responseObject.features.length === 1) {
+				if (responseObject.features[0].fid.has('gages')) {
+					this.router.navigate('#!streamflow-stats/gage/' + responseObject.features[0].attributes.STAID, {trigger : true});					
+				}
+				else {
+					var sortedFeature;
+					if (responseObject.features.length > 0) {
```
mbucknell: We should verify with @dblodgett-usgs, but I believe we no longer have to do this sorting because clicking on a huc only returns a single feature now rather than the upstream watersheds as well. So we should be able to remove the sorting. It doesn't do anything the way the code is writeen because you can't get to that point in the code unless there is only one feature.

dblodgett-usgs:
We no longer need the sorting...

mbucknell:
There's just a couple of things to rework. The rest looks good to go!

PR #284, opened 2015-08-18
watersmart-550; initial commit for task

mamcderm-usgs:
@mbucknell - Just a first pass to see if the approach is ok.  If so, is this "enough" or should I extend this approach to the project and data detail pages.  Also, I'm not sure whether or not Dave will want the error message displayed as I have it now (which use to show up in an alert), so I may need to remove that.


mbucknell:
Looks good to me. It's a question for Dave as to whether to extend to the project and data detail pages but I would say yes from the JIRA description.

mamcderm-usgs:
I'll go ahead and add handle the project and data detail pages the same way.

PR #288, opened 2015-08-21
Watersmart 208

mamcderm-usgs:


nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -196,7 +211,12 @@ NWC.util.DataSeriesStore = function () {
 				endOfMonth = Date.create(dayMetDateStr).utc().daysInMonth();
 				monthDateStr = dayMetDateStr;
 			}
-			monthlyAccumulation = saferAdd(monthlyAccumulation, dayMetValue);
+			//ignore a missing value at the end of the month
+			if (!dayMetValue && dayMetDay === endOfMonth) {
+			}
```
mbucknell: Can you refactor this if so that there isn't a blank clause after the if?

nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -231,6 +251,114 @@ NWC.util.DataSeriesStore = function () {
...
+				//if first time through but not beginning of year is true
+				else if (0 == etaMonths) {
+					endOfMonth = undefined;
+				}
+			}
+			//if first time through but not beginning of year is true skip until beginning of year			
+			if (undefined != endOfMonth) {
+				//ignore a missing value at the end of the month
+				if (!dayMetValue && dayMetDay === endOfMonth) {
+				}
```
mbucknell: If you really want to keep the if this way, I think it would be clearer if you put the comment in the if statement block.

nwc/src/test/javascript/specs/DataSeriesStore_spec.js
```diff
@@ -57,6 +58,23 @@ describe('DataSeriesStore', function(){
...
+            var dayMetIndex = DataSeriesStore.getIndexOfColumnNamed('dayMet');
+            var secondYearsAccumulatedDaymet = dss.yearly.data[etaOffsetInMonths][dayMetIndex];
+            expect(secondYearsAccumulatedDaymet).toBe(sum);
+        });
+        it('should place NaN in the eta field for a year having no eta records', function(){
+            //test before the daymet values start
+            var firstYearsAccumulatedEta = dss.yearly.data[etaIndex][etaIndex];
+            expect(isNaN(firstYearsAccumulatedEta)).toBe(true);
+        });
+    });
```
mbucknell: If you have time and its not too hard, it would be great to have a test for the leap year condition.

mbucknell:
Looks good! Merge when you want.

nwc/src/main/webapp/js/utils/dataSeriesStore.js
mamcderm-usgs: oops meant to come back to that and forgot...

nwc/src/main/webapp/js/utils/dataSeriesStore.js
mamcderm-usgs: I'm not understanding, are you suggesting to do the if a different way or the comments a different way or both?

mbucknell: I'm leaving it up to you. I prefer not having an empty if statement.

mamcderm-usgs: oh, I thought you were referring to a different if statement:-/  I'm planning to change that empty one.

PR #289, opened 2015-08-28
Watersmart 518 python skeleton and sitemap generation

cschroed-usgs:
Pulled the python scripts into this repo. Considered using a git submodule, but decided against it. Created a shell script to build the python virtual env, install dependencies, and generate the search engine files. Created a build profile that uses the exec plugin to place the generated search engine files in the war.


cschroed-usgs:
I don't need a review of the python code. I'm wondering what people think of this style of integration.

jiwalker-usgs:
The one thing I see here.  And I might just be missing something, but the use of the exec goal with ${outputDirectory} in there results in different artifact output based on how the build was run.  I'm guessing the plan would be something like `mvn exec package` which seems reasonable, but because the exec is fairly expensive (at least that was what I was assuming) it wouldn't happen with every build.  This would mean `mvn package` would likely be run at least some of the time.

I like the setup with the python runner from within maven, but I'd pull it out into a separate project that is built independently of the nwc project.  This results in version tracking between the two (which is a tricky problem), but we won't be running an expensive operation for every check in.

Just some quick thoughts.  Have a good weekend.

cschroed-usgs:
Thanks for the feedback. I'm realizing that I didn't provide enough context initially. Sorry about that. The search engine file generation produces 388MB of uncompressed skeletons, which compresses down to 180MB as a war. The sum of the normal build time and this generation process is about four minutes. I estimate less than 1 minute will be required to transfer the packaged war from a Jenkins machine to nexus (eros2eros ftw!). I figure 5 minutes is a reasonable time to wait for a CI job to finish, but is too long for local development. Since we want developers to get quick feedback when running locally, I implemented the search engine file generation in an opt-in fashion via build profiles. The search engine file generation only runs if the user activates the "generate-search-engine-files-mac-and-linux" build profile (and someday, "generate-search-engine-files-windows") and provides a valid geoserver url. For example:
```
mvn \
-Dgeoserver_for_search_engine_file_generation=http://cida.usgs.gov/nwc/geoserver/ \
-Pgenerate-search-engine-files-mac-and-linux \
clean package
```
^ That is the kind of build that Jenkins would run during snapshot and release builds, whereas, devs will be able to use good ole
```
mvn clean package
```
and the like locally during normal development tasks.

Ok, with all of that explained, do you still think the python portion should be pulled out into a separate project? It seemed like the advantage of the two-project idea was that you traded some bookkeeping for the ability to run the file generation separately. Since we can run the search file generation selectively with the setup I propose, and since the file generation is about 5 minutes, I can't figure out what the advantage of separate projects is. As you mentioned, the version-tracking issue between the two maven modules sounds annoying. There would be more repositories, poms, Jenkins jobs, and human steps to get fresh search engine content out to prod.

In the two-project scenario, If data is updated in geoserver or sciencebase then we must:
 * build a new generated-pages snapshot artifact
 * build a new nwc snapshot artifact 
 * deploy the nwc snapshot artifact
 * wait for it to pass review
 * perform a release build of the generated-pages artifact
 * manually remember-or-forget to update the reference to the generated-pages artifact version in our nwc pom
 * optionally, to make sure that the generated-pages artifact released correctly:
    * build a new nwc snapshot artifact 
    * deploy a new nwc snapshot artifact to mak
 * perform a release build of nwc
 * deploy the nwc release artifact

In the one-project scenario, If data is updated in geoserver or sciencebase then we must:
 * build a new nwc snapshot artifact 
 * deploy the nwc snapshot artifact
 * wait for it to pass review
 * perform a release build of nwc
 * deploy the nwc release artifact

Now /I/ might be missing something :) I'm curious! Let me know!

cschroed-usgs:
@jiwalker-usgs Thanks for chatting in person. I'm going to merge this to get the proof-of-concept working. I'd like to move to a multi-maven-module setup like the one you described after.

PR #290, opened 2015-08-31
Watersmart 545

mbucknell:
This is an initial start on adding a Config model which will contain (eventually) all application constants rather than sprinkling them throughout the code. Started with huc12 information. We will be shortly adding huc08 to the watershed budget page. At that time we can add a huc08 property to the watershed object along with the appropriate metadata.


nwc/src/main/webapp/js/init.js
```diff
@@ -48,6 +48,7 @@ $(document).ready(function() {
 	NWC.templates.registerHelpers();
 
 	$.when(loadTemplates, loadPartials).always(function() {
+		NWC.config = new NWC.model.Config();
```
mbucknell: Wasn't sure whether I should create this here or in the router. Another question is should this be put in the global namespace or passed in to views as needed.

mbucknell:
@mamcderm-usgs , if you have time, please review this, both for things I could do better but also getting in mind that we will be adding the ability to choose between huc08s and huc12s in the near future. I have further work to do on this to add additional configuration information that is already in the app.

nwc/src/main/webapp/js/init.js
mamcderm-usgs: Unfortunately, I don't really have the expertise to have a strong opinion on this.  It kind of seems like it should be passed in as needed.  Might be good to have input from Phethala on this.

mamcderm-usgs:
@mbucknell - I'm going to spend some more time stepping through this because my knowledge on models is minimal and rusty, and I want to understand the concepts better.  Anyway, it looks good to merge and didn't want to hold you up any more than I have.

mbucknell:
I'm adding some more config, so that at least all of the SOS constants and the streamflow gage data constants are in the config. The Config model is really just a convenient holder for this information. For the work that we will be doing next, we will be adding a huc08 DataSourceModel to the watershed property. The idea is that we will determine which config to use by parsing the hucID to see if it's an 8 or a 12 on the waterbudget data page. Keep this in mind as you look at the code more closely.

nwc/src/main/webapp/js/init.js
mbucknell: I'm now leaning on the side of creating this in the Config.js and only having the object have global visibility. No one needs to be creating new Config models. Plus if we ever use something like requirejs for dynamic module loading, we can then pass this around as a dependancy instead of global scope.

mbucknell:
Two significant changes. One is that all SOS configuration has been moved to the NWC.config model. I now create the NWC.config object in Config.js. All model definitions are no longer on the global namespace since nothing else should use these models.

mbucknell:
@mamcderm-usgs , I'm going to go ahead and merge. I will be leaving this ticket in implementing however as I may come back to it later in the sprint to move more things to the config model.

PR #292, opened 2015-09-01
Watersmart 518 python skeleton and sitemap generation

cschroed-usgs:
This is the first rough cut of serving the skeleton pages to searchbots. There is definitely room to improve. Let me know what you think!


nwc/src/main/java/gov/usgs/cida/nwc/util/PrettyUglyUrlMapper.java
```diff
@@ -0,0 +1,124 @@
+package gov.usgs.cida.nwc.util;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.charset.Charset;
+import java.util.List;
+import java.util.Locale;
+import org.apache.http.NameValuePair;
```
cschroed-usgs: These two imports are the only reason I brought in org.apache.httpcomponents.httpclient. Is that legit?

nwc/src/main/java/gov/usgs/cida/nwc/util/PrettyUglyPair.java
```diff
@@ -0,0 +1,11 @@
+package gov.usgs.cida.nwc.util;
```
cschroed-usgs: This should probably be in the test package -- that is the only place it is used.

nwc/src/main/java/gov/usgs/cida/nwc/util/PrettyUglyUrlMapper.java
```diff
@@ -0,0 +1,124 @@
...
+		URIBuilder uriBuilder = new URIBuilder(pretty);
+		String fragment = pretty.getFragment();
+		if (null != fragment && !fragment.isEmpty()) {
+			if (fragment.startsWith(BANG)) {
+				//move the content of the escaped fragment param
+				//to the fragment. Exclude the initial "!"
+				uriBuilder.addParameter(SEARCHBOT_ESCAPED_FRAGMENT_PARAM_NAME, fragment.substring(1));
+				uriBuilder.setFragment(null);
+			} else {
+				throw new IllegalArgumentException(
```
cschroed-usgs: Not sure this contract makes sense -- I throw an exception if there's a fragment that isn't started by a "!", but if there is no fragment, then I pass the same url back to the caller silently. Should I throw exceptions in both cases, or pass the same url back to the caller in both cases?

nwc/src/main/webapp/WEB-INF/web.xml
```diff
@@ -46,17 +46,24 @@
...
+    </servlet-mapping>
+    <servlet-mapping>
         <servlet-name>WPSProxy</servlet-name>
-         <url-pattern>/proxywps/*</url-pattern>
-    </servlet-mapping>	
-
-</web-app>
\ No newline at end of file
+        <url-pattern>/proxywps/*</url-pattern>
+    </servlet-mapping>
```
cschroed-usgs: Apparently there was a re-formatting. My bad. The following is the only thing that matters:

nwc/src/test/java/gov/usgs/cida/nwc/util/PrettyUglyUrlMapperTest.java
```diff
@@ -0,0 +1,92 @@
...
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import static org.junit.Assert.*;
+
+public class PrettyUglyUrlMapperTest {
+
+	public static final List<PrettyUglyPair> prettyUglyPairs = Arrays.asList(
```
cschroed-usgs: These are slightly modified versions of the example mappings on Google's spec.

cschroed-usgs:
I should note that this pull request does not include functionality to serve the sitemap xml files because the files are available by default at urls like <context-path>/sitemap.xml.

nwc/pom.xml
```diff
@@ -290,11 +294,25 @@
             <groupId>javax.servlet</groupId>
             <artifactId>jstl</artifactId>
         </dependency>
-
+	<dependency>
+		<groupId>org.mockito</groupId>
+		<artifactId>mockito-core</artifactId>
```
jiwalker-usgs: Do you want to change this scope to test?  I think you mention in a comment moving stuff to a test package.

cschroed-usgs: Good catch! Will do!

cschroed-usgs:
One thing to keep in mind -- I need to generalize this skeleton serving functionality to multiple projects. Some re-design will have to happen there. I'm thinking that the logic for converting a pretty url to a resource name will change from project to project, but the stuff that relies on the Google spec will stay the same.

cschroed-usgs:
In other words, the SkeletonPageServlet will have to become more general. Maybe the SkeletonPageServlet's constructor should accept an object that implement a PrettyUrlToResourceName interface. Each project would implement their own and specify it as an init param to SearchBotUglyUrlFilter in the web.xml. The filter would instantiate the parameterized class name and pass the instance to the SkeletonPageServlet's constructor. The servlet would use the PrettyUrlToResourceName instance to perform project-specific resource locating.

nwc/src/main/java/gov/usgs/cida/nwc/servlet/SkeletonPageServlet.java
```diff
@@ -0,0 +1,88 @@
...
+	 * and includes:
+	 *	* non-context path 
+	 *	* the query string
+	 *	* fragment
+	 * @param request
+	 * @return 
+	 */
+	String getUrlWithoutContextPath(String fullUrl, String contextPath){
+		if(null == contextPath || contextPath.isEmpty()){
+			throw new IllegalArgumentException("Received empty context path. Context path must be non-empty");
```
jiwalker-usgs: I'm not sure whether we can assume that we have a contextPath.  We don't do it much here, but it is a reasonable practice to deploy apps on the ROOT context.  This would cause problems in this case.

nwc/src/main/java/gov/usgs/cida/nwc/util/PrettyUglyUrlMapper.java
jiwalker-usgs: It is probably fine, maybe a bit of overkill, but better than writing it again.

nwc/src/main/java/gov/usgs/cida/nwc/util/PrettyUglyUrlMapper.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.http.client.utils.URLEncodedUtils;
+
+/**
+ * This class performs bidirectional mapping between ugly and pretty urls as
+ * specified in the Google specification for Making AJAX Applications Crawlable:
+ * https://developers.google.com/webmasters/ajax-crawling/docs/specification?hl=en
+ * 
+ */
+public class PrettyUglyUrlMapper {
+	static final String SEARCHBOT_ESCAPED_FRAGMENT_PARAM_NAME = "_escaped_fragment_";
```
jiwalker-usgs: This is defined twice, you might be able to find a common place for it.

nwc/src/test/java/gov/usgs/cida/nwc/util/PrettyUglyUrlMapperTest.java
```diff
@@ -0,0 +1,92 @@
...
+		//pretty url is a simple value
+		new PrettyUglyPair("/path#!hashfragment", "/path?_escaped_fragment_=hashfragment"),
+		//pretty url has ampersands and equals signs in the fragment:
+		new PrettyUglyPair("/#!key1=value1&key2=value2", "/?_escaped_fragment_=key1%3Dvalue1%26key2%3Dvalue2"),
+		//urls have query params besides the searchbot escaped fragment param
+		new PrettyUglyPair("/path?queryparams#!hashfragment", "/path?queryparams&_escaped_fragment_=hashfragment")
+	);
+	
+	public PrettyUglyUrlMapperTest() {
```
jiwalker-usgs: Just a suggestion, I usually don't include constructor , @Befores, @Afters if they aren't doing anything.

jiwalker-usgs:
It might be useful to look at some old code we had for dealing with peculiarities with URLs.  It is a bit dated, so might not fit with all the weirdness we've introduced since, but it has some golden oldies in there.  https://github.com/USGS-CIDA/tomcat-utilities

nwc/src/main/java/gov/usgs/cida/nwc/servlet/SkeletonPageServlet.java
cschroed-usgs: Yeah, I was being lazy. Thanks for bringing this up. I will address this.

nwc/src/test/java/gov/usgs/cida/nwc/util/PrettyUglyUrlMapperTest.java
cschroed-usgs: That's a good idea. I will do that.

cschroed-usgs:
Thanks for the suggestion. I'll check those tomcat utils out.

nwc/src/main/java/gov/usgs/cida/nwc/util/PrettyUglyUrlMapper.java
cschroed-usgs: Cool. Thanks!

nwc/src/main/java/gov/usgs/cida/nwc/util/PrettyUglyUrlMapper.java
cschroed-usgs: Done!

nwc/src/main/java/gov/usgs/cida/nwc/util/PrettyUglyUrlMapper.java
cschroed-usgs: I switched this to pass the same url back to the caller in both cases.

cschroed-usgs:
@jiwalker-usgs, I addressed all your comments so fat. The tomcat-utils repo didn't have anything I could use, but it's good to know it's there. Any other comments?

cschroed-usgs:
I'm going to defer the generalization refactoring that I rambled about above. Will address in a separate pull request.

jiwalker-usgs:
:+1: good to go

cschroed-usgs:
Thanks!

PR #299, opened 2015-09-03
Fixing sitemap validation errors.

cschroed-usgs:


nwc/src/main/python/templates/sitemap/index.xml
```diff
@@ -1,11 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
```
cschroed-usgs: @mbucknell , in case you work on sitemap.xml stuff in the future, it's worth knowing that you can't have a url element in a sitemapindex element. So even if there's only 1 url in the category, you've got to make a new file like home.xml above, which is silly, but apparently silly is the Way to Do It (TM) :smile:

PR #300, opened 2015-09-04
Watersmart 546

mamcderm-usgs:
I got the jira number wrong on this, it should be watersmart-547.


nwc/src/main/webapp/js/model/Config.js
```diff
@@ -63,6 +63,28 @@ NWC.model = NWC.model || {};
 								downloadMetadata: 'Data derived by sampling the SSEBop Actual Evapotranspiration dataset to NHD+\nVersion II 12-digit Hydrologic Unit Code Watersheds using the Geo Data Portal.\nhttp://cida.usgs.gov/thredds/catalog.html?dataset=cida.usgs.gov/ssebopeta/monthly\nhttp://cida.usgs.gov/gdp/ http://www.horizon-systems.com/NHDPlus/NHDPlusV2_home.php'
 							})
 						}
+					}),
+					huc8 : new DataSourceModel({
+						layerName : 'nationalwbdsnapshot', // get from mapUtils
+						namespace : 'NHDPlusHUCs', // getfrom mapUtils
```
mbucknell: I think the namespace and layerName should be WBD:huc08. @dblodgett-usgs can you verify that this is the layer to use for huc8 selection?

nwc/src/main/webapp/js/model/Config.js
```diff
@@ -63,6 +63,28 @@ NWC.model = NWC.model || {};
...
+						variables : {
+							dayMet : new SosVariable({
+								observedProperty: 'MEAN_prcp',
```
mbucknell: Should have had you look in WATERSMART-529 for these parameters. observedProperty should be 'prcp',  dataset should be HUC08_data, HUC08_daymet.nc.

nwc/src/main/webapp/js/model/Config.js
```diff
@@ -63,6 +63,28 @@ NWC.model = NWC.model || {};
...
+								propertyLongName: 'Area Weighted Mean Precipitation',
+								units: NWC.util.Units.metric.normalizedWater.daily,
+								dataset: 'HUC8_data',
+								fileName: 'HUC8_daymet.nc',
+								downloadMetadata: 'Data derived by sampling the DayMet precipitation variable to NHD+ Version II\n8-digit Hydrologic Unit Code Watersheds using the Geo Data Portal.\nhttp://daymet.ornl.gov/ http://cida.usgs.gov/gdp/\nhttp://www.horizon-systems.com/NHDPlus/NHDPlusV2_home.php'
+							}),
+							eta : new SosVariable({
+								observedProperty: 'MEAN_et',
```
mbucknell: observedProperty should be et, dataset should be HUC08_data, and fileName should be HUC08_eta.nc

nwc/src/main/webapp/js/utils/mapUtils.js
```diff
@@ -154,18 +154,25 @@ NWC.util.mapUtils = (function () {
 				fill: false
 			});
 
-		if (huc12s.length === 1) {
+		var hucProperty;
+		if (hucs[0].length === 8) {
```
mbucknell: We might consider putting the hucProperty in the Config too and then pass it in as a parameter.

nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -31,9 +31,14 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
 	 *	@prop {String} hucId - Previously selected watershed
 	 */
 	initialize : function(options) {
-		var watershedConfig = NWC.config.get('watershed').huc12.attributes;
+		var watershedHuc8Config = NWC.config.get('watershed').huc8.attributes;
```
mbucknell: It would be great to generalize this more by getting the 'watershed' config and then iterate over the properties to create the layers, etc. This way we can add additional hucs (maybe huc04's next) to the config and the code should just adapt.

nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -82,33 +95,98 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
 				fillColor: '#FF9900',
 				fillOpacity: 0.4
 			});
+			var namespace;
+			var layerName;
+			//these are actually the same but could change?
```
mbucknell: See comment above, because they are not the same.

nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -82,33 +95,98 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
...
+			if (options.hucId.length === 8) {
```
mbucknell: Consider adding utility functions for isHuc8 and isHuc12. I suspect they will be used in the ticket you are working on.

nwc/src/main/webapp/templates/waterbudget.html
```diff
@@ -29,9 +29,11 @@
 		<div class="row">
 			<div class="spacer col-xs-12">
 				{{> mapControls }}
-				<a id="toggle-huc-layer" class="btn btn-primary" data-toggle="button">Turn Watersheds <span id="toggle-huc-layer-span">On</span></a>
-				<span>Turn on/off map-display of <a href="http://www.horizon-systems.com/NHDPlus/NHDPlusV2_documentation.php" target="_blank">Watershed Boundary Dataset</a> <i class="fa fa-external-link"></i> for which water budget information is available.</span>
-
+				<div id="huc-controls-group" class="btn-group">
+					<a type="button" id="huc8-layer" class="btn btn-primary">Turn Huc8 Watersheds <span id="toggle-huc8-layer-span">On</span></a>
```
mbucknell: The ticket specified that this should be a select control rather than two radio buttons.

mbucknell:
There are some issues that will need to be fixed before completing this. You can merge if it makes things easier.

nwc/src/main/webapp/js/model/Config.js
dblodgett-usgs: I believe so. This is what is in the geoserver config project: 
{"name": "huc08", 
                        "workspace": "WBD", 
                        "style": "polygon",
                        "native_crs": "EPSG:4269",
                        "declared_crs": "EPSG:900913"}

cschroed-usgs:
@mamcderm-usgs could you please hold off on merging this until we get to a better spot with the search engine work? Normally I'd say merge early, merge often, but the search engine work needs to get propagated to production in order for us to verify that Google is indexing as expected, so keeping master free of incomplete features is a temporary priority.

cschroed-usgs:
In the mean time it's still ok to address Mary's suggestions and push your results to your fork so that she can see your progress in this pull request's diff. It's all good as long as you talk with me before hitting "Merge Pull Request."

mamcderm-usgs:
@mbucknell - this is ready to review again with changes based on previous comments.

nwc/src/main/webapp/js/view/HucCountyMapView.js
```diff
@@ -10,6 +10,7 @@ NWC.view.HucCountyMapView = NWC.view.BaseView.extend({
 	 *
 	 * @constructor
 	 * @param {Object} options
+	 *  @prop {String} hucId - Id of the huc for which information should be shown.
```
mbucknell: I believe this should be "huc" rather than "hucId" in the comment.

nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -15,13 +15,16 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
 	templateName : 'waterbudget',
 
 	events: {
-		'click #toggle-huc-layer' : 'toggleHucVisibility'
+		'change select' : 'toggleLayer'
```
mbucknell: Consider naming this something other than toggleLayer, perhaps selectLayer.

nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -51,21 +59,31 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
...
+		    	Object.keys(watershedConfig, function(key, value) {
+		    		if (actualFeature.attributes[value.attributes.property]) {
+						huc = actualFeature.attributes[value.attributes.property];
+		    		};
+		    	});
+		    	
 				if (Object.has(options, 'hucId')) {
-					this.router.navigate('#!waterbudget/comparehucs/' + options.hucId + '/' + huc12, {trigger : true});
+					if (options.hucId === huc) {
+						this.showWarningDialog('The same watershed has been selected for comparison. Please select a different watershed');
```
mbucknell: Good catch! Didn't think about this condition.

nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -31,18 +34,23 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
...
-
-		this.hucLayer = NWC.util.mapUtils.createHucLayer(watershedConfig.namespace, watershedConfig.layerName, {
-			visibility : false
-		});
+		var self = this;
+		var watershedConfig = NWC.config.get('watershed');
+				
+		var index = 0;
+		Object.keys(watershedConfig, function(key, value) {
+    		self.hucLayers[index] = NWC.util.mapUtils.createHucLayer(value.attributes.namespace, value.attributes.layerName, {
```
mbucknell: This should be indented.

nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -31,18 +34,23 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
...
 	initialize : function(options) {
-		var watershedConfig = NWC.config.get('watershed').huc12.attributes;
-
-		this.hucLayer = NWC.util.mapUtils.createHucLayer(watershedConfig.namespace, watershedConfig.layerName, {
-			visibility : false
-		});
+		var self = this;
+		var watershedConfig = NWC.config.get('watershed');
+				
+		var index = 0;
```
mbucknell: You don't need to do use an index. You can form the arrays by using 'arrayVar.push(arrayElement).

nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -51,21 +59,31 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
...
+				var huc;				
+		    	Object.keys(watershedConfig, function(key, value) {
+		    		if (actualFeature.attributes[value.attributes.property]) {
+						huc = actualFeature.attributes[value.attributes.property];
+		    		};
+		    	});
+		    	
 				if (Object.has(options, 'hucId')) {
-					this.router.navigate('#!waterbudget/comparehucs/' + options.hucId + '/' + huc12, {trigger : true});
+					if (options.hucId === huc) {
```
mbucknell: There should also be a check for huc null since that could happen. Probably just show an alert.

mbucknell:
Looks good. You should at least fix the indenting before merging. It makes the code very difficult to read. There were problems in more than the place I pointed out.

PR #301, opened 2015-09-04
Switching to search files bundled in external jar.

cschroed-usgs:


cschroed-usgs:
Again, [the pom](https://github.com/USGS-CIDA/nwc/pull/301/files?diff=unified#diff-378c1e19ef6e619f6f9bee9101e6b44a) is probably the only interesting thing in this pull request.

cschroed-usgs:
Merging so I can get started on integrating the two modules.

jiwalker-usgs:
:+1: So, if you could refresh my memory, how does the Filter know how to get to the files in the search-engine-files jar?

cschroed-usgs:
First, to avoid potential confusion, the filter only applies to searchbot ugly url (?_unescaped_param_) requests. The sitemaps are retrieved via the webjar META-INF/resources trick.

For the searchbot ugly url requests, we currently hardcode the resource name in the servlet that the filter chain delegates to. That hardcoding is bad because it is so project-specific. I'm working on a separate feature branch which moves the searchbot filter, servlet, and util classes into their own project-independent maven module. When that's done the filter will accept a project-specific class name as an init param in the web.xml. Each project can write their own tiny class that knows the path to the skeleton page resources in the search engine files jar.

jiwalker-usgs:
Yes, it was the skeleton I was referring to.  This sounds good to me, just clearing it up. :shipit:

cschroed-usgs:
Sweet. I'm happy to explain. Thanks!

cschroed-usgs:
Auugh -- it should say ?\_escaped_fragment_ my bad.

PR #305, opened 2015-09-14
Watersmart 546

mamcderm-usgs:
@mbucknell - I found the problem with the duplicate layers, which was that I was creating duplicate layers....go figure.  I'm not sure why this wasn't happening before I added the huc8 layer.  Regardless, would you take a look and see if I am handling the best way.

### DIFF
--- a/nwc/src/main/webapp/js/view/WaterBudgetMapView.js
+++ b/nwc/src/main/webapp/js/view/WaterBudgetMapView.js
@@ -18,13 +18,6 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
 		'change select' : 'selectLayer'
 	},
 
-	context : {
-		hucs : [{value : "none", display : "none"}],
-		warningModalTitle : 'Warning'
-	},
-	
-	hucLayers : [],
-
 	/**
 	 * @constructs
 	 * @param {Object} options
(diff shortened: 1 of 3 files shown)


nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -136,13 +150,12 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
 		var self = this;
 		var layer = this.model.get('watershedLayer');
 
-		this.hucLayersIndex.forEach(function(el, index) {
-			if (el === layer) {
-				self.hucLayers[index].setVisibility(true)
-			}
+		this.hucLayers.map(function(n) {
+			if (layer === n.propertyId)
```
mbucknell: You could make this if a one liner: n.layer.setVisibility(layer === n.propertyId);

mbucknell:
Before we added the huc 8's the layer creation was handled just a little differently. The changes look good to me!

mbucknell:
After thinking about this a little more, I don't think we should be reusing the watershedLayer. We should create a new one and replace the existing one. I'm not sure why we are saving this anyways. It isn't needed to restore the page and the data page should be fetching the huc feature itself and not depending on it being in the model.

mamcderm-usgs:
Yes, that makes sense.  Unfortunately, I'm not seeing how this.hucLayers is being persisted.  I would have thought it would get deleted in removeCurrentView(), but I don't even see it in the debugger.   I'm guessing it must be getting saved "indirectly" through the the map model but that even seems like a stretch.  I could explicitly delete it in the featureInfoHandler before navigating to the huc data page but that doesn't seem right.  Can you tell how/why it's getting saved?

mbucknell:
I'm afraid you'll probably have to do some deep debugging to see what is happening. I'm wondering if it isn't coming from the previous page, somehow.

mbucknell:
You also mention that you don't see it in the debugger. Make sure you follow the prototype property down. I have been fooled by this before.

PR #307, opened 2015-09-16
watersmart-557; initial commit for task to add gage layer to waterbud…

mamcderm-usgs:
…get map

### DIFF
--- a/nwc/src/main/webapp/js/model/WaterBudgetSelectMapModel.js
+++ b/nwc/src/main/webapp/js/model/WaterBudgetSelectMapModel.js
@@ -10,7 +10,8 @@ NWC.model = NWC.model || {};
 NWC.model.WaterBudgetSelectMapModel = NWC.model.BaseSelectMapModel.extend({
 	defaults : function() {
 		return $.extend({
-			watershedLayer : "none"
+			watershedLayer : "none",
+			gageLayerOn : false,
 		}, NWC.model.BaseSelectMapModel.prototype.defaults);
 	}
 });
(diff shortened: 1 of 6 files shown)


nwc/src/main/webapp/js/model/WaterBudgetSelectMapModel.js
```diff
@@ -10,7 +10,8 @@ NWC.model = NWC.model || {};
 NWC.model.WaterBudgetSelectMapModel = NWC.model.BaseSelectMapModel.extend({
 	defaults : function() {
 		return $.extend({
-			watershedLayer : "none"
+			watershedLayer : "none",
+			gageLayerOn : false,
```
mbucknell: Remove the trailing ','

nwc/src/main/webapp/js/view/BaseSelectMapView.js
```diff
@@ -195,6 +195,13 @@ NWC.view.BaseSelectMapView = NWC.view.BaseView.extend({
 			this.zoomBoxControl.deactivate();
 			this.selectControl.deactivate();
 		}
-	}
+	},
+
+	legendUrl : function(layer, style) {
```
mbucknell: Great idea to move this here!

nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -39,6 +41,32 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
...
+				tiled: true
+			},
+			{
+				isBaseLayer: false,
+				displayInLayerSwitcher: false,
+				visibility: false
+			}
+		);
+
+		_.map(this.hucLayers, function(val, key) {
```
mbucknell: The more appropriate underscore functionto use here would be _.each. _.map should be used when you want to transform an object/array into another array.

mbucknell:
Looks good! :+1:

PR #310, opened 2015-09-18
Fix to config

mbucknell:


mbucknell:
@mamcderm-usgs , this should fix the display of data for huc 12s. I'm going to go ahead and merge.

PR #311, opened 2015-09-21
Nwc 548

mbucknell:
Moved the constants and functions from CountyWaterUseProperties into Config.js. Created a NWC.model.WaterUseColleciton to hold these constant and values and initialize them for county water use during the Config model instantiation. 

Cleaned up waterUsageChart but passing in the config object and making the assumptions about its functions clearer with more descriptive comments.


mbucknell:
@mamcderm-usgs , please review when you have time.

mamcderm-usgs:
looks great to me :+1:

PR #314, opened 2015-10-01
Watersmart 559

mamcderm-usgs:


nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -136,15 +138,63 @@ NWC.util.DataSeriesStore = function () {
...
+		etaSeries = nameToSeriesMap.eta,
+		nwisStreamFlowDataSeries = nameToSeriesMap.nwisStreamFlowData,
+		nwisStreamFlowDataSeriesMatch = false;
+
+		//if we have a streamflow series, get the first entry that matches
+		//up with the earliest dayMet series entry.  may want to change this
+		//in the future to plot any streamflow values that occur before
+		//the dayMet entries.
+		if (nwisStreamFlowDataSeries) {
+			nwisStreamFlowDataSeries.data.each(function (nwisStreamFlowData) {
```
mamcderm-usgs: This would go outside of this function and only called once for all three series, but not even sure if this makes sense.

nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -136,15 +138,63 @@ NWC.util.DataSeriesStore = function () {
...
 			var dayMetDateStr = dayMetRow[0],
 			dayMetValue = dayMetRow[1],
 			dayMetDay = getDayNumberFromDateString(dayMetDateStr);
+			//currently, we are using the dayMet range to govern this method
+			//and therefore we need to sync the streamflow data with the first 
+			//daymet date or earliest daymet date, if streamflow data comes
+			//after the first daymet values. this whole thing probably needs
+			//to change though.
+			if (nwisStreamFlowDataSeriesMatch) {
+				if (nwisStreamFlowDataSeries.data[nwisStreamFlowDataIndex][0] > dayMetRow[0]) {
```
mamcderm-usgs: there is probably a better way to do this part too.  Dave said it was ok to "bound" the graph by dayMet values but it seems like the whole thing may need to be refactored to account for the streamflow being optional and the values possibly being outside the date range of dayMet.

nwc/src/main/webapp/js/model/Config.js
```diff
@@ -159,6 +182,9 @@ NWC.model = NWC.model || {};
 					color : '#f5833c'
 				}
 			]);
+			
+			var watershedGages = new NWC.model.WatershedGagesCollection();
+			watershedGages.fetch({reset : true});
```
mbucknell: This introduces a potential race condition. The "fetch" function returns a jqXHR object (essentially a deferred). I think it would be better to define initialize for WatershedGagesCollection, call fetch, and then wait until the deferred is resolved or rejected.

nwc/src/main/webapp/js/model/Config.js
```diff
@@ -170,6 +196,7 @@ NWC.model = NWC.model || {};
 						namespace : 'WBD',
 						property : 'huc12',
 						name : 'name',
+						watershedAcres : 'areaacres', //should this be in variables?
```
mbucknell: No. I think this is correct as it is property key just like "property" and "name" are for the WFS call.

nwc/src/main/webapp/js/model/Config.js
```diff
@@ -195,6 +222,7 @@ NWC.model = NWC.model || {};
 						namespace : 'WBD',
 						property : 'huc8',
 						name : 'name',
+						watershedAcres : 'areaacres', //should this be in variables?
```
mbucknell: I don't think this should be defined for huc 8s.

mbucknell:
@mamcderm-usgs , I'm going to have to come back to this tomorrow which I will do first thing.

nwc/src/main/webapp/js/model/Config.js
mamcderm-usgs: I really wrestled with this one.  I could not figure out how to reference the data coming in from the file on initialize.  It seems like you would have to pass the file location as a parameter on the constructor, but nothing I tried worked.  So, I switched to using the parse which get's fired on fetch.  It didn't seem like initialize got fired on fetch.  Anyway, do you know of any examples of doing this in the cida code base somewhere (couldn't find any comparable examples online).

mbucknell: Can you call fetch from initialize?

nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -136,15 +138,63 @@ NWC.util.DataSeriesStore = function () {
...
+		nwisStreamFlowDataSeries = nameToSeriesMap.nwisStreamFlowData,
+		nwisStreamFlowDataSeriesMatch = false;
+
+		//if we have a streamflow series, get the first entry that matches
+		//up with the earliest dayMet series entry.  may want to change this
+		//in the future to plot any streamflow values that occur before
+		//the dayMet entries.
+		if (nwisStreamFlowDataSeries) {
+			nwisStreamFlowDataSeries.data.each(function (nwisStreamFlowData) {
+				if (nwisStreamFlowData[0] < dayMetSeriesStartDate) {
```
mbucknell: Rather than using the column index number, we should be using columnIndices.date (etc).

nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -136,15 +138,63 @@ NWC.util.DataSeriesStore = function () {
...
+					nwisStreamFlowDataIndex++;
```
mbucknell: One of the parameters to Sugar's each is index (see http://sugarjs.com/api/Array/each). Consider using that rather than incrementing an index.

nwc/src/main/webapp/js/model/Config.js
mamcderm-usgs: hmmm, backbone documentation says " If you define an initialize function, it will be invoked when the collection is created."  This seems to imply that initialize would get fired by fetch, so it would need to be outside of initialize?  However, when I have the fetch outside, it does not seem to fire initialize (passing in data as parameter to constructor does).  I may just be doing something wrong though....

nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -136,15 +138,63 @@ NWC.util.DataSeriesStore = function () {
...
+				}
+				else if (nwisStreamFlowData[0] > dayMetSeriesStartDate){
```
mbucknell: Can you explain the purpose of this else if clause?

nwc/src/main/webapp/js/model/Config.js
mbucknell: The method fetch does not create the collection it merely adds models to an existing collection.

mamcderm-usgs: Ok, I can see that initialize is getting fired on instantiation, but how do you reference the collection inside initialize to fetch the models?

mbucknell: Use this --- this.fetch()

nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -136,15 +138,63 @@ NWC.util.DataSeriesStore = function () {
...
+							nwisStreamFlowDataIndex++;	
+						}
+					});
+				}
+				else {
+					nwisStreamFlowDataSeriesMatch = true;
+					return false;					
+				}
+			});
+		}
```
mbucknell: I think we can simplify things. I may be wrong, but I think what we want to do is limit the nwisStreamFlowDataSeries to only include dates in the same range as the daymet series. Can we use Sugar (or Underscore) helpers? I think we could use the filter function and pass it a function that returns true only for dates that are within the date range of the daymet series.

nwc/src/main/webapp/js/model/Config.js
mamcderm-usgs: That's what I thought, but doesn't seem to load anything...I must be doing something stupid.

nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -136,15 +138,63 @@ NWC.util.DataSeriesStore = function () {
...
 		dayMetSeries.data.each(function (dayMetRow) {
 			var dayMetDateStr = dayMetRow[0],
 			dayMetValue = dayMetRow[1],
 			dayMetDay = getDayNumberFromDateString(dayMetDateStr);
+			//currently, we are using the dayMet range to govern this method
+			//and therefore we need to sync the streamflow data with the first 
+			//daymet date or earliest daymet date, if streamflow data comes
+			//after the first daymet values. this whole thing probably needs
+			//to change though.
+			if (nwisStreamFlowDataSeriesMatch) {
```
mbucknell: If we do something like I suggested above, I think we can simplify it to add the NaN for missing dates.

nwc/src/main/webapp/js/model/Config.js
mbucknell: So it doesn't make the ajax call?

nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -359,8 +418,10 @@ NWC.util.DataSeriesStore = function () {
 	* DataSeries objects
 	*/
 	self.updateHucSeries = function (nameToSeriesMap) {
-		this.eta = nameToSeriesMap.eta;
-		this.dayMet = nameToSeriesMap.dayMet;
+//it doesn't look like we need these...remove?
```
mbucknell: I think I would create a method to limit the nwisStreamFlowData to the date range of daymet before calling the update*HucSeries. That way you only have to do it once.

nwc/src/main/webapp/js/model/Config.js
mamcderm-usgs: It does make the call, but doesn't seem to get the response and/or add models to the collection.  After the fetch(), the models arrays is empty:
		initialize : function () {
			this.fetch();
			this.lookupByHucId = {};
			_.each(this, function(watershed) {
				this.lookupByHucId[watershed.hucId] = watershed;
			}, this);
			return this;
		},

nwc/src/main/webapp/js/view/HucInsetMapView.js
```diff
@@ -21,14 +21,26 @@ NWC.view = NWC.view || {};
...
+			}
+			else {
+				watershedConfig = NWC.config.getWatershed(options.hucId);
+			}
+
 			var baseLayer = NWC.util.mapUtils.createWorldStreetMapLayer();
 			var mapControls = [new OpenLayers.Control.Zoom(), new OpenLayers.Control.Navigation()];
 			var hucLoadedDeferred = $.Deferred();
 
+			//do we want to put gage Id on view?
```
mbucknell: I don't see why. Probably don't need to put hucId on it either.

nwc/src/main/webapp/js/view/WaterbudgetPlotView.js
```diff
@@ -56,40 +70,106 @@ NWC.view = NWC.view || {};
...
-						//@todo - setup app level error handling
-						var errorMessage = 'Error retrieving time series data for ' + this.sourceId;
-						alert(errorMessage);
-						d.reject(errorMessage);
-					}
-				});
+				
+				if (sourceId === 'nwisStreamFlowData') {					
+					var startDate = '1838-01-01';  //this is the earliest date for all dv data
+					var endDate = '';  //use today? or, just leave blank?
```
mbucknell: Can we leave out the date parameters?

nwc/src/main/webapp/js/view/WaterbudgetPlotView.js
```diff
@@ -56,40 +70,106 @@ NWC.view = NWC.view || {};
...
+								}
+								row.push(parseDateStr($(this).attr('dateTime')));
+								row.push(value);
+								dataTable.push(row);
+							});
+
+							if (dataTable.length === 0) {
+								var errorMessage = 'No data available to plot ' + self.sourceId;
+								alert(errorMessage);
+//								d.reject(errorMessage);  //do we want to reject?
```
mbucknell: I would say no and probably don't alert either. You could still put a message in the console.

mbucknell:
@mamcderm-usgs, I think it would be best to have a function to limit the nwis data to the daymet range in dataSeriesStore. I will pull down your code and see if I can figure out what is happening with the fetch in Config or come up with a different way of doing this.

There will eventually need to be some commenting and code cleanup. In dataSeriesStore, clearing state your assumptions.

nwc/src/main/webapp/js/view/HucInsetMapView.js
```diff
@@ -55,6 +69,21 @@ NWC.view = NWC.view || {};
...
+			
+			var gageId = options.gageId ? options.gageId : null;
+			if (gageId) {
+				this.streamflowGageConfig = NWC.config.get('streamflow').gage.attributes;
+				this.gageLayer = NWC.util.mapUtils.createGageFeatureLayer(
+						this.streamflowGageConfig.namespace,
+						this.streamflowGageConfig.layerName,
+						gageId);
+				//might need to change the gage marker symbol?
+				//also might need other gage info?
```
mamcderm-usgs: @mbucknell - what is the proper way to change the gage marker to a blue dot here?  Do you modify the layer directly or do you do it "through a marker"?

mbucknell: We could probably streamline this by adding some styling to the gageLayer, but for now let's just create the marker as is done in StreamflowStatsGageDataView.

nwc/src/main/webapp/js/view/WaterbudgetPlotView.js
mamcderm-usgs: I played with this and went through the waterservices documentation but I couldn't figure out how to get values back without providing at least a startDate...but it wouldn't surprise me if there is a way.

mamcderm-usgs:
@mbucknell - I think I got pretty close to what you suggested with the dataseries, at least the graphs are working.  I think I hit most, if not all of your other comments as well, so hopefully this is getting close.  Appreciate if you would take another look when you get a chance.  I'm going to move on to the tests for now.  Thank you for all your help!

nwc/src/main/webapp/js/utils/dataSeriesStore.js
```diff
@@ -116,6 +117,20 @@ NWC.util.DataSeriesStore = function () {
...
 
+	//if we have a streamflow series, trim any entries that may be present
+	//that precede the first dayMet series entry and follow the last dayMet
+	//series entry.	
+	self.trimNwisStreamFlowData = function (nwisSeries) {
+		var dayMetSeriesStartDate = this.dayMet.data[0][0];
+		var total = _.size(self.dayMet.data);
+		var dayMetSeriesEndDate = self.dayMet.data[_.size(self.dayMet.data)-1][0];
+
+		self.nwisData = _.filter(nwisSeries.data, function (entry) {
```
mbucknell: Rather than create a new property, can we just assign the results of _.filter back to self.nwisStreamFlowData? If we do that, I would not pass in the parameter nwisSeries.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -37,19 +39,30 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
...
+		if (this.accumulated) {
+			this.context.showWaterUseButton = false;
+			this.context.showAccumulatedButton = false;
+		}
+		else {
+			//if huc_12 watershed and local view, show button for accumulated water budget
+			this.context.showAccumulatedButton = (this.hucId.length === 12);			
+			this.context.showWaterUseButton = true;			
+		}
+
```
mbucknell: Consider simpfying this if-else clause to two statements, something like 
this.context.showWaterUseButton = !this.accumulated;
this.context.showAccumulatedButton = (!this.accumulated) || (this.hucId.length === 12);

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -99,6 +133,16 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
 			});
 		}
 	},
+	
+	accumulatedPlotView : function () {
```
mbucknell: OCD comment: Can we change this to createAccumulatedPlotView? Functions should be a verb IMO.

nwc/src/main/webapp/js/view/WaterbudgetPlotView.js
```diff
@@ -39,13 +50,16 @@ NWC.view = NWC.view || {};
...
+		* accumulated type of view from the WaterBudgetHucDataView and there is an associated gage
+		* with the selected watershed. 
 		*
 		* @param {String} huc 12 digit identifier for the hydrologic unit
 		* @returns a resolved promise when both ETA and DAYMET data has been retrieved and the dataSeriesStore updated. If
 		*      either call fails the promise will be rejected with either one or two error messages. The datasSeriesStore object will
 		*      contain the data for any successful calls
 		*/
-		getHucData: function(huc) {
+		getHucData: function(huc, gage) {
```
mbucknell: Maybe we should just call this getPlotData since it potentially gets both huc and gage data.

nwc/src/main/webapp/js/view/WaterbudgetPlotView.js
```diff
@@ -39,13 +50,16 @@ NWC.view = NWC.view || {};
 
 		/**
 		* This makes a Web service call to get huc data and transform it into a data series object.
+		* This will also make a web service call to get streamflow data if it is created as an
+		* accumulated type of view from the WaterBudgetHucDataView and there is an associated gage
+		* with the selected watershed. 
 		*
 		* @param {String} huc 12 digit identifier for the hydrologic unit
 		* @returns a resolved promise when both ETA and DAYMET data has been retrieved and the dataSeriesStore updated. If
```
mbucknell: You should add the gage parameter as well.

nwc/src/main/webapp/js/view/WaterbudgetPlotView.js
```diff
@@ -56,40 +70,105 @@ NWC.view = NWC.view || {};
...
-						//@todo - setup app level error handling
-						var errorMessage = 'Error retrieving time series data for ' + this.sourceId;
-						alert(errorMessage);
-						d.reject(errorMessage);
-					}
-				});
+				
+				if (sourceId === 'nwisStreamFlowData') {					
+					var startDate = '1838-01-01';
+					var endDate = '';
```
mbucknell: This is not used and should be removed.

nwc/src/main/webapp/js/view/WaterbudgetPlotView.js
```diff
@@ -102,9 +181,10 @@ NWC.view = NWC.view || {};
...
 		},
 
-
 		/**
-		 * Update the plot with the current dataSeriesStore and model.
+		 *	Update the plot with the current dataSeriesStore and model.
+		 *	If this is an accumulated type of view from the WaterBudgetHucDataView and there is
+		 *	an associated gage with the selected watershed, then streamflow will also be plotted. 
 		 */
 		plotPTandETaData : function() {
```
mbucknell: Maybe we should just call this plotData since it now plots more the precipation and eta data.

mbucknell:
@mamcderm-usgs , it looks good!

PR #324, opened 2015-10-13
Watersmart 560

mamcderm-usgs:
@mbucknell - This seems to be working ok, but still considering ways to make the code more efficient.  Appreciate any suggestions.


nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -59,7 +59,60 @@ NWC.view.WaterBudgetHucDataView = NWC.view.BaseView.extend({
...
+			this.hucInsetMapView.hucFeatureLoadedPromise.done(function() {
+				self.plotView = new NWC.view.WaterbudgetPlotView({
+					accumulated : true,
+					hucId : self.hucId,
+					gageId : self.gageId,
+					el : self.$('#huc-plotview-div'),
+					model : self.hucPlotModel
+				});
+
+				/*	create the compare hucInsetmap view after the plot data has been retrieved
```
mbucknell: Can we have two distinct properties for watershedAcres in the WaterBudgetHucPlotModel? It really should be two so that we know what each represents. Also that way you don't have to wait to create the compareHucInsetMapView.

nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -23,11 +23,16 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
 	 * @constructs
 	 * @param {Object} options
 	 *	@prop {String} mapDiv
+	 *  @prop {Boolean} accumulated - true indicates accumulated, which indicates an
```
mbucknell: Fix indentation

nwc/src/test/javascript/specs/WaterBudgetHucDataView_spec.js
```diff
@@ -239,12 +242,55 @@ describe('Tests for NWC.WaterBudgetHucDataView', function() {
...
 			expect($compareHucsButton.prop('disabled')).toBe(true);
 			hucFeatureLoadedDeferred.resolve();
+			hucDataLoadedDeferred.resolve();
 			expect($countiesButton.prop('disabled')).toBe(true);
 			expect($accumulatedButton.prop('disabled')).toBe(true);
 			expect($compareHucsButton.prop('disabled')).toBe(false);
 		});
 	});
 
+	describe('Tests for view created with a compareHucId and accumulated watershed', function() {
```
mbucknell: Great tests!

mbucknell:
@mamcderm-usgs, probably should make another two properties for the watershedAcres. Otherwise, it looks good to go.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
mamcderm-usgs: I had thought about doing this but just wasn't sure.  It seems like I will need to make the HucInsetMap view "compare aware", I guess by passing it a flag then using that flag to set the appropriate watershedAcres model variable.  Is that what you were thinking or is there a simpler way to do this that I'm not seeing?

mbucknell: You'll have to make it "compare aware" in some fashion. Passing in a flag is probably simpliest.

PR #335, opened 2015-10-20
switch back to dev mode

dblodgett-usgs:


nwc/src/main/webapp/js/model/Config.js
```diff
@@ -184,7 +184,7 @@ NWC.model = NWC.model || {};
 
 			return {
 				featureToggles : {
-					enableAccumulatedWaterBudget : false
```
thongsav-usgs: May want to load this from a JNDI property from the webservice so you don't have to rebuild/deploy for a toggle.

dblodgett-usgs:
I don't disagree... but this is supposed to be a temporary toggle. Hopefully we won't have to switch it again.

PR #339, opened 2015-10-30
IE11 meta tag and prepare to re-release

dblodgett-usgs:
@mbucknell I verified that this shows up in the page that loads initially, is this what you would expect?


mbucknell:
Yes ... it's in the correct place.

PR #341, opened 2015-11-20
changes so we only show gages that have a drainage area within 85% of…

dblodgett-usgs:
… the huc area


nwc/src/main/webapp/js/view/WaterBudgetMapView.js
```diff
@@ -54,10 +54,10 @@ NWC.view.WaterBudgetMapView = NWC.view.BaseSelectMapView.extend({
 			CONFIG.endpoint.geoserver + 'NWC/wms',
 			{
 				LAYERS: gageConfig.namespace + ':' + gageConfig.layerName,
-				STYLES: 'blue_circle',
+				STYLES: 'gagesii_hucComp',
 				format: 'image/png',
 				transparent: true,
-				tiled: true
+				tiled: false
```
dblodgett-usgs: I don't know why, but if I used tiled true, the style would work.

PR #343, opened 2016-03-21
Watersmart 585

mbucknell:
When showing huc12s we are now always showing both the accumulated and local watersheds on both the local and accumulated pages. They are styled differently depending on what kind of page is shown. Added a legend to distinguish the two polygons and added additional information.

This does not complete the ticket, but the rest of the changes are related to layout.

### DIFF
--- a/nwc/src/main/webapp/js/view/HucInsetMapView.js
+++ b/nwc/src/main/webapp/js/view/HucInsetMapView.js
@@ -35,7 +35,8 @@ NWC.view = NWC.view || {};
 
 			this.context = {
 				hucId : hucId,
-				gageId : gageId
+				gageId : gageId,
+				isHuc12 : hucId.length === 12
 			};
 
 			var hucLoadedDeferred = $.Deferred();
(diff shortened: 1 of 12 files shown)


nwc/src/main/webapp/js/view/HucInsetMapView.js
```diff
@@ -7,16 +7,23 @@ NWC.view = NWC.view || {};
...
+		strokeOpacity : 1,
+		fill: false
+	};
+
+
 	NWC.view.HucInsetMapView = NWC.view.BaseView.extend({
 
 		templateName : 'hucInsetMap',
 
-		// This  is resolved once the huc feature layer has been loaded and the hucLayer property
```
mbucknell: These should not be defined at this level. When this package is instantiated a second time, those promises will already be defined. Also, now making a single promise visible on the view's object which is resolved when all layers have been retrieved.

nwc/src/main/webapp/js/view/WaterBudgetHucDataView.js
```diff
@@ -1,278 +1,217 @@
-/*jslint browser: true*/
```
mbucknell: Refactored the initialize for this module. Not sure why git is marking all lines as changed.

nwc/src/test/javascript/specs/WaterBudgetHucDataView_spec.js
```diff
@@ -1,361 +1,395 @@
-/*jslint browser: true */
```
mbucknell: Not sure why git thinks this changed. Most of the test have not been changed. I had to move the spy for HucInsetMapView within the describe for each suite of tests. For the compare, since there are two instantiations of HucInsetMapView, I had to mock the featureLoadedPromise differently.

mbucknell:
The layout adjustments have been made. I believe I have implemented all of the changes in the ticket.

mamcderm-usgs:
Thanks for the comments :+1:

PR #349, opened 2016-04-01
texas API version change

jkreft-usgs:
looks like the 1.0 version of the API is dead


mbucknell:
That's mighty unfriendly to just have the version go away!

dblodgett-usgs:
Thanks for the catch!

PR #364, opened 2016-07-19
Adding commercial to industrial.

dblodgett-usgs:


dblodgett-usgs:
@mbucknell This is one of the things that was asked for in the process of figuring out the HUC12 water use.

PR #371, opened 2016-08-09
Nwc 615

mbucknell:


nwc/src/main/webapp/vendorlibs/jquery.download.js
```diff
@@ -12,14 +12,21 @@
 jQuery.download = function(url, data, method){
```
mbucknell: Changed this so rather than making a query string from an object of name/value pairs and then parsing it back into name and values that are inserted into the DOM, it will if data is a query string, parse this into name/value pairs that then can easily be put in the DOM. This solved an issue where the value string valued contained '=' character as the cql_filter value does.

nwc/src/main/webapp/js/view/WaterbudgetDownloadUpstreamView.js
```diff
@@ -110,6 +111,38 @@ NWC.view = NWC.view || {};
...
 			this._downloadUpstreamData('HUC12_data/HUC12_daymet.nc', 'prcp');
+		},
+
+		downloadShapefile : function() {
+			var $loadingMsg = this.$('.download-msg');
+			var $loadingAlert = this.$('.download-msg-container');
+			var $loadingIndicator = this.$('.download-loading-indicator');
+			var upstreamHucs = this.model.get('upstreamHucs');
+
+			if (upstreamHucs.length > CONFIG.maxUpstreamHucsForShapefileDownload) {
```
ayan-usgs: I'm wondering if it's worth putting a test in for this behavior when a shapefile can't be downloaded... if it's a lot of work,  I'll leave it to your discretion whether to add one or not.

ayan-usgs:
Looks good. Merge at your discretion.

nwc/src/main/webapp/js/view/WaterbudgetDownloadUpstreamView.js
mbucknell: When I added this view in a recent ticket, I opted not to write tests for it. Much of the hard work of the previous two download buttons is done in another module. Its a lot harder to set up DOM for tests in NWC due to the way we structured the code. I'm going to opt out of this. I feel bad about it though!

PR #376, opened 2016-08-10
Setting partial months/years for nwis or modeled streamflow data to NaN

mbucknell:


dblodgett-usgs:
Wow... youreally did just have to switch it! Glad this was basically already there.

